PR #4, opened 2020-01-16
include db queries

countless-integers:
adjust the upstream code to work with the newer agent version.

**I think I got the unit of time wrong though.**


countless-integers:
Closed in favour of #5

PR #5, opened 2020-01-16
feat: collect executed DB queries

arkaitzgarro:
- Create a data collector that registers DB query executions.
- Add a variable to limit the number of spans created.

### DIFF
--- a/src/Events/Span.php
+++ b/src/Events/Span.php
@@ -1,0 +1,40 @@
+<?php
+namespace AG\ElasticApmLaravel\Events;
+
+use PhilKra\Events\Span as BaseSpan;
+
+class Span extends BaseSpan
+{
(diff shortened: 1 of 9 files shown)


src/Events/Span.php
```diff
@@ -0,0 +1,40 @@
+<?php
+namespace AG\ElasticApmLaravel\Events;
+
+use PhilKra\Events\Span as BaseSpan;
+
+class Span extends BaseSpan
```
arkaitzgarro, line 6: I created a [PR](https://github.com/philkra/elastic-apm-php-agent/pull/121) to avoid this situation. Until then, there is no other way of setting the start time of a Span.

countless-integers (approved):

PR #7, opened 2020-01-23
fix: measured times for application startup

arkaitzgarro:
Split the startup time in two:
- Application load time, usually taken by Composer, file reading, etc
- Framework load, measured by Laravel's events `booting` and `booted`.

Add the time that it takes for a request to be handled, measured by the time between `RouteMatched` and `RequestHandled` events. I suppose this also includes middlewares.

Add the query name: query type + target tables.

Screenshot 2020-01-24 at 09 50 27

### DIFF
--- a/src/Collectors/DBQueryCollector.php
+++ b/src/Collectors/DBQueryCollector.php
@@ -1,9 +1,14 @@
 <?php
 namespace AG\ElasticApmLaravel\Collectors;
 
+use Exception;
+
 use Illuminate\Events\Dispatcher;
 use Illuminate\Support\Collection;
 use Illuminate\Database\Events\QueryExecuted;
+
+use Jasny\DB\MySQL\QuerySplitter;
+
 use AG\ElasticApmLaravel\Collectors\TimelineDataCollector;
 use AG\ElasticApmLaravel\Collectors\Interfaces\DataCollectorInterface;
 
@@ -12,15 +17,6 @@ use AG\ElasticApmLaravel\Collectors\Interfaces\DataCollectorInterface;
  */
 class DBQueryCollector extends TimelineDataCollector implements DataCollectorInterface
 {
-    protected $request_start_time;
-
-    public function __construct($request_start_time)
-    {
-        parent::__construct();
-
-        $this->request_start_time = $request_start_time;
-    }
-
     public function onQueryExecutedEvent(QueryExecuted $query): void
     {
 
@@ -34,7 +30,7 @@ class DBQueryCollector extends TimelineDataCollector implements DataCollectorInt
         $end_time = $start_time + $query->time / 1000;
 
         $query = [
-            'name' => 'Eloquent Query',
+            'name' => $this->getQueryName($query->sql),
             'type' => 'db.mysql.query',
             'action' => 'query',
             'start' => $start_time,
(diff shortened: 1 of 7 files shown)


arkaitzgarro:
@countless-integers im going to clean up the code a bit, and extract parts from the `ServiceProvider` to separate classes and start writing some tests.

src/ServiceProvider.php
```diff
@@ -121,4 +137,28 @@ protected function getAppConfig(): array
 
         return $config;
     }
+
+    protected function getController()
```
countless-integers:
```suggestion
    protected function getController(): ?string
```

src/Collectors/DBQueryCollector.php
```diff
@@ -61,4 +57,23 @@ public static function getName(): string
...
         return 'query-collector';
     }
+
+    private function getQueryName(string $sql): string
+    {
+        $fallback = 'Eloquent Query';
+
+        try {
+            $query_type = QuerySplitter::getQueryType($sql);
+            $tables = QuerySplitter::splitTables($sql);
```
countless-integers, line 83: Interesting package, lots of regexes. Builds are failing, but it seems to do what we need it to do.

src/ServiceProvider.php
```diff
@@ -55,14 +57,28 @@ protected function registerAgent(): void
...
+    protected function listenForEvents(): void
     {
+        $start_time = $this->app['request']->server('REQUEST_TIME_FLOAT') ?? microtime(true);
         $timeline_collector = $this->app->make(Agent::class)->getCollector(TimelineDataCollector::getName());
+        
+        // Application and Laravel startup times
+        $timeline_collector->startMeasure('app_boot', 'app', 'boot', 'App boot', $start_time);
+        $this->app->booting(function () use ($timeline_collector) {
+            $timeline_collector->stopMeasure('app_boot');
+            $timeline_collector->startMeasure('laravel_boot', 'laravel', 'boot', 'Laravel boot');
```
countless-integers: What's the difference between `app_boot` and `laravel_boot`?

src/Collectors/DBQueryCollector.php
```diff
@@ -61,4 +57,23 @@ public static function getName(): string
...
+
+            if (isset($query_type) && is_array($tables)) {
+                // Query type and tables
+                return $query_type . ' ' . join(', ', array_values($tables));
+            }
+
+            return $fallback;
+        } catch (Exception $e) {
+            return $fallback;
```
countless-integers, line 92: You could also keep the catch block empty and move the `return $fallback` after the whole try-catch block to same 1 line of code (#crazy-gains).

src/Collectors/DBQueryCollector.php
arkaitzgarro: Yeap, but it does the job, and it's also really fast.

src/ServiceProvider.php
arkaitzgarro: We might need a better naming indeed. If I understand correctly how Laravel works, the difference between both of them is as follows:
- `app_boot`:  the time for loading Composer autoloader file, create Laravel application instance, create the container and register service providers.
- `laravel_boot`: load environment variables, register facades, resolve the Kernel, register routes, register middlewares...

src/Collectors/DBQueryCollector.php
arkaitzgarro: I never liked empty `catch` blocks 😅

countless-integers: Fair enough.

countless-integers (approved):
🚢

PR #8, opened 2020-01-24
chore: respect configuration when registering services

countless-integers:
`LARAVEL_START` is not defined when running tests with Codeception, so I'd apply a hammer solution to use that config option to prevent loading services when we set it.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -29,9 +29,6 @@ class ServiceProvider extends BaseServiceProvider
      */
     public function register(): void
     {
-        if (!config('elastic-apm-laravel.active')) {
-            return;
-        }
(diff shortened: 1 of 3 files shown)


src/ServiceProvider.php
```diff
@@ -29,6 +29,9 @@ public function boot(): void
      */
     public function register(): void
     {
+        if (!config('elastic-apm-laravel.active')) {
```
arkaitzgarro: I'm afraid this will make the middleware fail because it expects the Agent to be defined.

Not registering the listeners should be enough. It will still create the transaction, but nothing else will be collected and also the agent will not send anything to Elastic. Let me make the change for you.

arkaitzgarro (approved):

PR #9, opened 2020-01-27
fix: disable agent and middleware based on config

arkaitzgarro:
If clients are not enabling APM, let's not register the agent nor the middleware. If it's enabled, we are responsible for registering the agent and the middleware at the very top of the list, hoping for more accurate measure times.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,14 +17,7 @@ Add the ServiceProvider class to the providers array in `config/app.php`:
 ],
 ```
 
-And register the [global middleware](https://laravel.com/docs/5.5/middleware#global-middleware) that will be called on every request in `app/Http/Kernel.php`:
-
-```php
-protected $middleware = [
-    // ... more middlewares
-    \AG\ElasticApmLaravel\Middleware\RecordTransaction::class,
-];
-```
+From here, we will take care of everything based on your configuration. The agent and the middleware will be registered, and transactions will be send to Elastic.
 
 ## Agent configuration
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -39,8 +32,6 @@ The following environment variables are supported in the default configuration:
 |APM_SECRETTOKEN    | Secret token, if required. |
 |APM_QUERYLOG       | `true` or `false` defaults to 'true'. Set to `false` to completely disable query logging, or to `auto` if you would like to use the threshold feature. |
 |APM_THRESHOLD      | Query threshold in milliseconds, defaults to `200`. If a query takes longer then 200ms, we enable the query log. Make sure you set `APM_QUERYLOG=auto`. |
-|APM_BACKTRACEDEPTH | Defaults to `25`. Depth of backtrace in query span. |
-|APM_RENDERSOURCE   | Defaults to `true`. Include source code in query span. |
```
arkaitzgarro, line 43: I removed these two properties because we are not using them yet.

countless-integers: It's still in https://github.com/arkaitzgarro/elastic-apm-laravel/blob/71bf40baeb0ff03edeb8aedafbab8eafc1ba43c7/config/elastic-apm-laravel.php .

arkaitzgarro: `APM_BACKTRACEDEPTH` is used [by the agent](https://github.com/philkra/elastic-apm-php-agent/blob/1fdf7cf342d62603b7668d99e1fab7300c8ae2cf/src/Agent.php#L186), we should not remove it.

countless-integers (approved):

PR #11, opened 2020-02-08
Job tracking

cykirsch:
Add job tracking via middleware available in Laravel 6+ as well as via job events.

This splits framework init into its own collector so that it doesn't log duplicates between HTTP and Job.

### DIFF
--- a/src/Collectors/JobCollector.php
+++ b/src/Collectors/JobCollector.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace AG\ElasticApmLaravel\Collectors;
+
+use AG\ElasticApmLaravel\Agent;
+use AG\ElasticApmLaravel\Collectors\Interfaces\DataCollectorInterface;
+use Illuminate\Foundation\Application;
+use Illuminate\Queue\Events\JobProcessed;
+use Illuminate\Queue\Events\JobProcessing;
+
+/**
+ * Collects info about the http request process
+ */
+class JobCollector extends TimelineDataCollector implements DataCollectorInterface
+{
+    protected $app;
+    protected $agent;
+
+    public function __construct(Application $app, Agent $agent, float $request_start_time)
+    {
+        $this->app = $app;
+        parent::__construct($request_start_time);
+
+        $this->agent = $agent;
+        $this->registerEventListeners();
+    }
+
+    public static function getName(): string
+    {
+        return 'job-collector';
+    }
+
+    protected function registerEventListeners(): void
+    {
+        $this->app->events->listen(JobProcessing::class, function () {
+            $this->startMeasure('job_processing', 'job', 'processing', 'Job processing');
+        });
+
+        $this->app->events->listen(JobProcessed::class, function () {
+            if ($this->hasStartedMeasure('job_processing')) {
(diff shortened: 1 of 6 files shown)


src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public static function getName(): string
+    {
+        return 'job-collector';
+    }
+
+    protected function registerEventListeners(): void
+    {
+        $this->app->events->listen(JobProcessing::class, function () {
```
cykirsch, line 35: I'm actually thinking that this should be opposite. The order of execution is JobProcessing, Middleware, JobProcessed. So as it is now the Job processing span extends a few milliseconds beyond the end of the transaction.

I did it this way to intentionally be similar to the HTTP Request, but since it's not fully accurate, I'm wondering if you'd prefer it switched--the transaction would be handled by the events and the middleware would just provide an additional span. This would also have the benefit of automatic transactions for pre-L6 installs.

cykirsch:
@arkaitzgarro Any thoughts on this one so far?

cykirsch:
FWIW I'm about done with a re-implementation doing the opposite as described above; if it works out I could do a PR for that and then you can give feedback on whichever is your preferred approach.

arkaitzgarro:
Do you have an example of how both implementations visually look like in ELastic APM?

cykirsch:
> Do you have an example of how both implementations visually look like in ELastic APM?

Pretty subtle difference in the demo, but you can see right at the end where the transaction ends in comparison to the "Job processing" span.

Transaction started/stopped in the event:
Screen Shot 2020-02-11 at 11 57 33 AM

Transaction started/stopped in the middleware:
Screen Shot 2020-02-11 at 12 02 09 PM

cykirsch:
Building your own demo like the above is trivial. Sharing an artisan command below, but you can do the same thing in a controller to see the job-within-a-request.

```
<?php

namespace App\Console\Commands;

use App\Jobs\ApmJobDemoJob;
use Illuminate\Console\Command;
use Illuminate\Contracts\Bus\Dispatcher;

class ApmJobDemo extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'apm:job-demo';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Demo job tracking in Elastic APM';

    /**
     * Execute the console command.
     *
     * @param Dispatcher $dispatcher
     */
    public function handle(Dispatcher $dispatcher)
    {
        $dispatcher->dispatch(new ApmJobDemoJob());
    }
}
```

```
<?php

namespace App\Jobs;

use AG\ElasticApmLaravel\Jobs\Middleware\RecordTransaction;
use App\Models\Eloquent\User;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Queue\SerializesModels;

class ApmJobDemoJob implements ShouldQueue
{
    use SerializesModels;

    public function handle()
    {
        $user = User::first();
    }

    public function middleware()
    {
        return [
            app(RecordTransaction::class),
        ];
    }
}
```

arkaitzgarro:
Closing in favour of #18

PR #15, opened 2020-02-08
chore: setup PHP Coding Standards

arkaitzgarro:
Adhere to [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) and [Symfony](https://symfony.com/doc/current/contributing/code/standards.html) coding standards.

PR #17, opened 2020-02-10
Event-based custom span collection

cykirsch:
Create StartMeasuring and StopMeasuring events to allow collection of spans manually instrumented throughout the application.

### DIFF
--- a/src/Collectors/SpanCollector.php
+++ b/src/Collectors/SpanCollector.php
@@ -2,7 +2,6 @@
 
 namespace AG\ElasticApmLaravel\Collectors;
 
-use AG\ElasticApmLaravel\Agent;
 use AG\ElasticApmLaravel\Collectors\Interfaces\DataCollectorInterface;
 use AG\ElasticApmLaravel\Events\StartMeasuring;
 use AG\ElasticApmLaravel\Events\StopMeasuring;
@@ -29,13 +28,12 @@ class SpanCollector extends TimelineDataCollector implements DataCollectorInterf
     protected function registerEventListeners(): void
     {
         $this->app->events->listen(StartMeasuring::class, function (StartMeasuring $event) {
-            // TODO: Throw exception or log warning if measure exists?
             $this->startMeasure(
                 $event->name,
                 $event->type,
                 $event->action,
                 $event->label,
-                $event->startTime,
+                $event->start_time,
             );
         });
 
--- a/src/Collectors/SpanCollector.php
+++ b/src/Collectors/SpanCollector.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace AG\ElasticApmLaravel\Collectors;
+
+use AG\ElasticApmLaravel\Agent;
+use AG\ElasticApmLaravel\Collectors\Interfaces\DataCollectorInterface;
+use AG\ElasticApmLaravel\Events\StartMeasuring;
+use AG\ElasticApmLaravel\Events\StopMeasuring;
+use Illuminate\Foundation\Application;
+
+/**
+ * Generic collector for spans measured manually throughout the app.
+ */
+class SpanCollector extends TimelineDataCollector implements DataCollectorInterface
+{
+    public function __construct(Application $app, float $request_start_time)
+    {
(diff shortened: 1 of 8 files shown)


arkaitzgarro:
Hey @cykirsch, thanks again for  your contribution!

I'm wondering if in that case we could reuse the existing `TimelineDataCollector` instead of adding an extra wrapper based on events. Something like:

```php
// src/Agent.php
// Collector for manual measurements throughout the app
$this->collectors->put(
    TimelineDataCollector::getName(),
    new TimelineDataCollector($app, $this->request_start_time)
):
```

```php
// src/ServiceProvider.php
$start_time = $this->app['request']->server('REQUEST_TIME_FLOAT') ?? microtime(true);
$this->app->instance(TimelineDataCollector::class, new TimelineDataCollector($this->app, $start_time));
```

```php
// src/your/file.php
$timelineCollector = $this->app->make(TimelineDataCollector::class); // app()->make or App::make()
$timelineCollector->startMeasure('my-custom-span', 'custom', 'measure', 'My custom span')

// do something amazing

$timelineCollector->stopMeasure('my-custom-span');
```

We could rename the TimelineCollector class to SpanCollector, if it makes it easier to understand the responsibility of the class.

cykirsch:
I ultimately did the events for some key reasons:

- follow the existing event->collector architecture so it's easy to understand
- very simple to fire from anywhere without being concerned about config
- do as little as possible under the hood when APM_ACTIVE is not set (fire event but don't listen)
- no change to TimelineDataCollector because so far it has not been used directly and is a good candidate for being an abstract class

If you're not fond of the events, I'd next suggest a service whose methods would each do `isActive` checks so that it still short circuits when configuration is off. This could have a facade wrapper on it for ease of use. This is the original approach I took but I was far happier when I realized how clean it became when I switched to events.

src/Collectors/SpanCollector.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace AG\ElasticApmLaravel\Collectors;
+
+use AG\ElasticApmLaravel\Agent;
```
arkaitzgarro: This class is not used here right?

src/Collectors/SpanCollector.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public static function getName(): string
+    {
+        return 'span-collector';
+    }
+
+    protected function registerEventListeners(): void
+    {
+        $this->app->events->listen(StartMeasuring::class, function (StartMeasuring $event) {
+            // TODO: Throw exception or log warning if measure exists?
```
arkaitzgarro: Could you maybe open an issue with that question? This looks like a responsibility for `TimelineDataCollector` class.

src/Events/StartMeasuring.php
```diff
@@ -0,0 +1,35 @@
...
+    public $type;
+
+    /** @var string|null */
+    public $action;
+
+    /** @var string|null */
+    public $label;
+
+    /** @var float|null */
+    public $startTime;
```
arkaitzgarro: Can you define variables as `snake_case`? I know we have a bit of a mess now with code standards, but I'm working on it https://github.com/arkaitzgarro/elastic-apm-laravel/pull/15.

arkaitzgarro (commented):
I get your point on events, and I don't dislike it. I was just thinking in a way to solve it without adding more classes, but it likes the way to go. As you mention, I think we should even create a facade for better user experience:

```php
// Names are hard, it could also be ApmCollector so it can be used in other frameworks
ApmLaravel::startMeasure('my-custom-span', 'custom', 'measure', 'My custom span');

// do something amazing

ApmLaravel::stopMeasure('my-custom-span');
```

src/Events/StartMeasuring.php
arkaitzgarro: Well, Symfony coding standards expect variables in `camelCase` and configuration in `snake_case`. It is basically the opposite of what we are doing 😅

arkaitzgarro (approved):
This is lovely, thanks 🙌

PR #18, opened 2020-02-11
Job tracking with transaction handled by events

cykirsch:
This is an alternative to https://github.com/arkaitzgarro/elastic-apm-laravel/pull/11.
They accomplish the same thing but the order of tracking is reversed. In the other one, the transaction is captured via job middleware and a span via events. However, the actual order of processing is such that the middleware happens between the events, and therefore the events should handle the transaction, hence this PR.


src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->app->events->listen(JobProcessed::class, function (JobProcessed $event) {
+            $transactionName = $this->getTransactionName($event);
+            $this->addMetadata($transactionName, $event->job);
+            $this->stopTransaction($transactionName);
+        });
+    }
+
+    protected function startTransaction(string $transactionName): Transaction
+    {
+        return $this->agent->startTransaction(
```
arkaitzgarro, line 46: Will this will start a Job transaction inside the Request transaction? I have to admit that I never tried to do that. I'm curious to see how does it look like.

cykirsch: I did a test around that and it got a little funky, but yes it's supposed to be possible. For example, if you do `dispatchNow()`on `Illuminate\Bus\Dispatcher` it runs the job within the same process so you'd start 2 transactions in that case.

I spent a little time trying to build up a $transactionStack so we can set the [parentTransaction](https://github.com/philkra/elastic-apm-php-agent/blob/master/docs/examples/parent-transactions.php) (the existing transactionStore doesn't seem conducive to simply finding the last transaction started). It did make it better but I think my top level transaction looked strange. I have those changes stashed so I can revive them if needed.

cykirsch: Controller with dispatchNow(), seems to ignore the job events and middleware entirely, everything falls under the controller transaction. I think that's fine and probably right.
Screen Shot 2020-02-11 at 12 26 25 PM

With dispatch() - the controller transaction is empty
Screen Shot 2020-02-11 at 12 29 37 PM

...and the job has everything
Screen Shot 2020-02-11 at 12 30 10 PM

**But if we set up parent transactions...**

Then the job still has everything
Screen Shot 2020-02-11 at 2 21 36 PM

but the parent (controller) has zero length, and you can't see it but it's 'generic' instead of 'HTTP' 
Screen Shot 2020-02-11 at 2 21 12 PM

My conclusion is that we should do parent transactions but try to diagnose the 0 transaction timespan and incorrect transaction type.

cykirsch: Just an update, I think I did figure out where this needs to go. The problem with the empty one is that the `send()` flushes the transactionStore. So we should only call `collectEvents($name)`, and get rid of `sendTransaction($name)`. But after that it fills out the top transaction but double counts every span:
Screen Shot 2020-02-11 at 4 51 35 PM

To fix that, I think the collectors need to store the latest open transaction name on the Measure. Then `collectEvents` would only gather the events for the given name. So each span would reference the innermost transaction as its parent.

I'd like to take a crack at that tomorrow if you think this is progressing in a way that you'll be comfortable with. It makes sense to me... 😂

cykirsch: @arkaitzgarro I did spend a good deal of time on this. The transaction stack basically works. But there's pretty significant issues with the time tracking. That mostly centers around the fact that the agent and timeline collector just take in the request time and stick to it. Since the span time is relative to the transaction, each transaction should actually get its own timestamp. This issue seems to flow all the way up to PhilKra though.

I tweaked times left and right until I got it quite close, but not all the way there. It does involve getting rid of REQUEST_TIME_FLOAT and LARAVEL_START because they get in the way, and also requires extending the Transaction object, not unlike LazySpan.

I don't have it in a clean state, most of this is from messing around in the vendor directory. But I could get it into a branch so it's documented at least. Depending what you think, it may be better to support jobs in a simpler state but leave parent transactions somewhat unsupported AKA "known issue". 

I will say they work pretty nice in Kibana - you can click to go see just the child portion, and then from the child there is a "View Full Trace" button to go back to the parent.

arkaitzgarro: @cykirsch I feel your pain with timing, that's also the reason I had to create the `LazySpan` class, because PhilKra events take as start time the moment when the instance was created. I also tried to extend the `Transaction` class, but didn't success that much since base properties are private. I even forked the `elastic-apm-php-agent` repo and start rewriting it, but it was too much...

I'd prefer indeed to start simple, without parent transactions at least to have a way to measure jobs, but knowing that it can be better and can be implemented in the future.

src/Collectors/HttpRequestCollector.php
```diff
@@ -79,4 +73,9 @@ protected function getController(): ?string
 
         return $controller;
     }
+
+    private function isHttpRequest(): bool
+    {
+        return false !== strpos($_SERVER['SCRIPT_NAME'], 'index.php');
```
countless-integers: Wouldn't [`php_sapi_name`](https://www.php.net/manual/en/function.php-sapi-name.php) or [`PHP_SAPI`](https://www.php.net/manual/en/reserved.constants.php#reserved.constants.core) do the same thing?

src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    protected function addMetadata($transactionName, $job): void
+    {
+        $this->agent->getTransaction($transactionName)->setMeta([
+            'type' => 'job'
+        ]);
+    }
+
+    protected function stopTransaction($transactionName) : void
```
countless-integers: Could you also add type-hinting to all the methods that don't have it?

```suggestion
    protected function stopTransaction(string $transactionName): void
```

src/Collectors/HttpRequestCollector.php
cykirsch: I suppose so. Based on https://github.com/philkra/elastic-apm-laravel/issues/19, I also went ahead and added the option to disable non-HTTP while leaving HTTP logging active.

cykirsch:
I appreciate the feedback, this is ready for 👀again.

cykirsch:
@arkaitzgarro @countless-integers Anything you'd like to see on this one that I could take care of before the weekend?

arkaitzgarro:
@cykirsch I will take a look later this afternoon.

README.md
```diff
@@ -79,20 +80,6 @@ php artisan vendor:publish --tag=config
 
 Once published, open the `config/elastic-apm-laravel.php` file and review the various settings.
 
-## Manual span tracking
```
arkaitzgarro, line 133: Any specific reason to remove this piece of documentation?

src/Collectors/HttpRequestCollector.php
```diff
@@ -12,22 +11,18 @@
...
+        return 'request-collector';
+    }
 
-        $this->app->booted(function () {
-            $this->startMeasure('route_matching', 'laravel', 'request', 'Route matching');
-            $this->stopMeasure('laravel_boot');
-        });
+    protected function registerEventListeners(): void
+    {
+        if ('cli' !== php_sapi_name()) {
```
arkaitzgarro: Maybe we could add this conditional to the `ServiceProvide`, to not add the collector at all if it's not a HTTP request?

src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,81 @@
...
+    protected function startTransaction(string $transaction_name): Transaction
+    {
+        return $this->agent->startTransaction(
+            $transaction_name,
+            [],
+            $this->request_start_time
+        );
+    }
+
+    protected function addMetadata($transaction_name, $job): void
```
arkaitzgarro: Looks like `$job` is not used inside the function. Please provide the type for `$transaction_name` argument.

README.md
```diff
@@ -60,7 +60,8 @@ The following environment variables are supported in the default configuration:
 
 | Variable          | Description |
 |-------------------|-------------|
-|APM_ACTIVE         | `true` or `false` defaults to `true`. If `false`, the agent will collect, but not send, transaction data. |
+|APM_ACTIVE         | `true` or `false` defaults to `true`. If `false`, the agent will collect, but not send, transaction data; span collection will also be disabled. |
+|APM_ACTIVE_CLI     | `true` or `false` defaults to `true`. If `false`, the agent will not collect or send transaction or span data for non-HTTP requests but HTTP requests will still follow APM_ACTIVE. |
```
arkaitzgarro: Maybe worth to mention that when `APM_ACTIVE` is `false`, non-HTTP requests will not be measured even if the value is `true`.

src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $this->agent->getTransaction($transaction_name)->setMeta([
+            'type' => 'job',
+        ]);
+    }
+
+    protected function stopTransaction($transaction_name): void
```
arkaitzgarro: Please provide the type for $transaction_name argument.

src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,81 @@
...
+        try {
+            // Stop the transaction and measure the time
+            $this->agent->stopTransaction($transaction_name);
+            $this->agent->collectEvents($transaction_name);
+        } catch (Throwable $t) {
+            Log::error($t->getMessage());
+        }
+    }
+
+    protected function getTransactionName($event)
```
arkaitzgarro: Please provide a return type for the method.

src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        return Arr::get($event->job->payload(), 'displayName', 'Default');
```
arkaitzgarro: Could `get_class($event->job)` be a better default value?

src/ServiceProvider.php
```diff
@@ -30,7 +30,10 @@ public function register(): void
         // Always available, even when inactive
         $this->registerFacades();
 
-        if (false === config('elastic-apm-laravel.active')) {
+        if (
+            false === config('elastic-apm-laravel.active')
+            || ('cli' === php_sapi_name() && false === config('elastic-apm-laravel.cli.active'))
```
arkaitzgarro: If I'm reading this conditional correctly, this will never register the Agent, middlewares, and collectors when `elastic-apm-laravel.active` is `true` but we are executing via CLI and `elastic-apm-laravel.cli.active` is false.

arkaitzgarro (commented):
I went through the code and left some comments. I'm going to test it in a Laravel 5 application.

src/Collectors/JobCollector.php
arkaitzgarro: Will not `$event->job->getName()` achieve what you need?

README.md
cykirsch: Yes, it's included under the "Collectors" section higher up now.

src/Collectors/HttpRequestCollector.php
cykirsch: Yep, but it will be in the Agent, under registerCollectors()

src/Collectors/JobCollector.php
cykirsch: I left `$job` intentionally but will document why--this doesn't do much by default so it provides a bit more info for someone who might extend this method. Let me know if you'd still like it removed.

arkaitzgarro: I'm afraid the linter will complain. Maybe we could actually use the variable to add metadata, like the number of retries of the queue name.

src/Collectors/JobCollector.php
cykirsch: `get_class($event->job)` is `Illuminate\Queue\Jobs\SyncJob`. Yeah, that would be a better default.
I don't think `$event->job->getName()` is useful to monitoring at all, in my current test it's `Illuminate\Queue\CallQueuedHandler@call`

But I poked around a bit more and found `$event->job->resolveName()` does what I want under the hood. Thanks for questioning this one!

src/ServiceProvider.php
cykirsch: If I read your comment correctly, that's right.

If `.active` is `false`, it will not register, regardless of request or `cli.active`.

If `.active` is `true` and we are on HTTP, it will register.

If `.active` is `true`, we are on CLI, and `cli.active` is false, it will not register.

If `.active` is `true`, we are on CLI, and `cli.active` is true, it will register.

src/Collectors/JobCollector.php
cykirsch: It hasn't yet, I don't think php-cs-fixer has that rule as an option. PHPMD does though. https://phpmd.org/rules/unusedcode.html

cykirsch: I went ahead and made a change to log the result just like HTTP does. I didn't know what else to do so I went with 200 for success and 400 for failure since it's familiar.

src/Jobs/Middleware/RecordTransaction.php
```diff
@@ -0,0 +1,35 @@
...
+    public function handle($job, $next)
+    {
+        if (false === config('elastic-apm-laravel.active') || false === config('elastic-apm-laravel.cli.active')) {
+            return $next($job);
+        }
+
+        /** @var Agent */
+        $agent = app(Agent::class);
+        /** @var JobCollector */
+        $collector = $agent->getCollector(JobCollector::getName());
```
cykirsch: @arkaitzgarro I'm not sure what to do here. My issue is that I can no longer call `getName` because it's not static. I see 2 obvious solutions

- I believe I saw @countless-integers mention using a constant instead? That would help here.
- Maybe I should just use the facade instead. I thought it was nice to keep these measurements within the job but maybe there's no reason they need to be. I think I'll go that route but I'll leave this comment anyway in case you have other opinions.

arkaitzgarro: I'm also thinking that we could ditch the `getName` method, and use the class of the collector as a name. That could also prevent any problem if a custom collector uses the same name as we do with default collectors.

src/Collectors/JobCollector.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->app->events->listen(JobProcessed::class, function (JobProcessed $event) {
+            $transaction_name = $this->getTransactionName($event);
+            $this->stopTransaction($transaction_name);
+            $this->setTransactionResult($transaction_name, 200);
+        });
+
+        $this->app->events->listen(JobFailed::class, function (JobFailed $event) {
+            $transaction_name = $this->getTransactionName($event);
+            $this->stopTransaction($transaction_name);
+            $this->setTransactionResult($transaction_name, 400);
```
arkaitzgarro: If a job failed is usually because of an exception right? If that's the case, I would rather prefer a 500 status code to make it really visible. 4XX codes can regular app validation response. @countless-integers what do you think?

countless-integers: Same. I'd keep 4XX errors for client generated errors only and in this case it's our app's wiring that causes the issue.

src/ServiceProvider.php
```diff
@@ -130,4 +130,10 @@ protected function getAppConfig(): array
 
         return $config;
     }
+
+    private function isAgentDisabled(): bool
+    {
+        return false === config('elastic-apm-laravel.active')
```
cykirsch, line 136: @arkaitzgarro @countless-integers This is duplicated logic with the ApmCollectorService, and the job middleware could safely share it as well. Any advice on making it more available? It could go in the service for access via facade, but I'm not sure if something else makes more sense.
What makes me hesitate on that idea is that the facade isn't going to load until it's needed, but if we call it from the ServiceProvider it will always load.

countless-integers: I have no problem with just checking the config value tbh. It's usually the first thing we want to check before loading anything. Could have a facade method, but then it would mean two ways of checking the same thing.

arkaitzgarro: I'm also ok for now. We could extend [philkra's config class](https://github.com/philkra/elastic-apm-php-agent/blob/master/src/Helper/Config.php) with some handy methods, but definetely not in this PR.

PR #21, opened 2020-02-11
Log warning when starting existing measure or stopping non-existing measure

### LINKED ISSUE #19: Add protection in startMeasure if it is already started
Likely in `TimelineDataCollector`. I think an exception might be extreme but a log could helpful. Potentially:

```
    public function startMeasure(
        string $name,
        string $type = 'request',
        string $action = null,
        string $label = null,
        float $start_time = null
    ): void {
        if ($this->hasStartedMeasure($name) {
            // Throw Exception?
            // Log Warning?
            return;
        }

        $start = $start_time ?? microtime(true);
        $this->started_measures->put($name, [
            'label' => $label ?: $name,
            'start' => $start - $this->request_start_time,
            'type' => $type,
            'action' => $action,
        ]);
    }
```

cykirsch:
Fixes #19

### DIFF
--- a/src/Collectors/HttpRequestCollector.php
+++ b/src/Collectors/HttpRequestCollector.php
@@ -36,23 +36,17 @@ class HttpRequestCollector extends TimelineDataCollector implements DataCollecto
 
         $this->app->booted(function () {
             $this->startMeasure('route_matching', 'laravel', 'request', 'Route matching');
-            if ($this->hasStartedMeasure('laravel_boot')) {
-                $this->stopMeasure('laravel_boot');
-            }
+            $this->stopMeasure('laravel_boot');
         });
 
         // Time between route resolution and request handled
         $this->app->events->listen(RouteMatched::class, function () {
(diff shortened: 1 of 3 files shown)


arkaitzgarro:
Now that we log possible mistakes, I think we should also remove the guards from collectors:
- https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Collectors/SpanCollector.php#L41
- https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Collectors/HttpRequestCollector.php#L39
- https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Collectors/HttpRequestCollector.php#L47
- https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Collectors/HttpRequestCollector.php#L53

cykirsch:
Good call, pushing soon.

arkaitzgarro (approved):

PR #23, opened 2020-02-11
chore: unit tests for Timeline Collector

arkaitzgarro:
Add unit tests for `TimelineDataCollector` class. Added mockery as a dependency to mock Laravel's `Log` facade and a package to include deprecated `assertArraySubset` PHPUnit method.


tests/Collectors/TimelineDataCollectorTest.php
```diff
@@ -0,0 +1,93 @@
...
+        ], $event);
+    }
+
+    public function testStartMeasureDuplication()
+    {
+        Log::shouldReceive('warning')
+            ->once()
+            ->with("Did not start measure 'test-measure' because it's already started.");
+
+        $this->collector->startMeasure('test-measure');
```
countless-integers: `'test-measure'` could be a const :).

tests/Collectors/TimelineDataCollectorTest.php
```diff
@@ -0,0 +1,93 @@
...
+    public function testCollectorName()
+    {
+        $this->assertEquals('timeline', TimelineDataCollector::getName());
+    }
+
+    public function testStartMeasure()
+    {
+        $this->collector->startMeasure(
+            'test-measure',
+            'request',
```
countless-integers, line 40: I wish PHP would introduce named parameters...

tests/Collectors/TimelineDataCollectorTest.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace AG\Tests\Collectors;
+
+use AG\ElasticApmLaravel\Collectors\TimelineDataCollector;
+use DMS\PHPUnitExtensions\ArraySubset\Assert;
+use Illuminate\Support\Facades\Log;
+
+class TimelineDataCollectorTest extends \Codeception\Test\Unit
```
countless-integers: `Unit` class could be imported like all others.

src/Collectors/TimelineDataCollector.php
```diff
@@ -86,9 +86,9 @@ public function addMeasure(
         string $label,
         float $start,
         float $end,
-        string $type = 'request',
-        string $action = 'request',
-        array $context = []
+        ?string $type = 'request',
```
countless-integers: Why make it nullable? What should happen if we explicitly pass `null` as a param (aka missing test case)?

countless-integers (changes requested):

src/Collectors/TimelineDataCollector.php
arkaitzgarro: I actually had to make it nullable because tests where failing 😀

When starting a measure and providing only a name, `$action` and `$label` properties will be null. When stopping the measure, we call `addMeasure` method with `null` values, while the method expect strings. Removing it will make tests fail:

```
Test  tests/Collectors/TimelineDataCollectorTest.php:testStartMeasureDuplication

  [TypeError] Argument 5 passed to AG\ElasticApmLaravel\Collectors\TimelineDataCollector::addMeasure() must be of the type string, null given, called in /home/vagrant/elastic-apm-laravel/src/Collectors/TimelineDataCollector.php on line 73
```

arkaitzgarro: But `$request` has always a default value, no need to make it nullable.

countless-integers: oh... I see it now.

countless-integers (approved):

PR #24, opened 2020-02-13
Add collector to the Agent

### LINKED ISSUE #20: Extendable collectors
My core issue is that I want to hook into the agent to track jobs. My preferred solution is #18, but I'm also ok with #11. 

But if not that, then there are a couple things I think should be done to allow better extensibility:

1. Split up HttpCollector so the framework stuff is done in its own collector. See the above PRs for a sample FrameworkCollector implementation. This is needed so that when I add my own collector for jobs, it doesn't have a "Route matching" span that never ends. I supposed you could just go with the `isHttpRequest` to not start the route_matching but I do think it's a bit awkward to have the boot stuff in http--may be just me
2. Since the Agent is a singleton, I can pick that up anywhere, but since `$collectors` isn't public, I can't just add my own. Either a `getCollectors` method or a public property would be fine.

Even if one of my job PRs is merged, step 2 may still be useful.

cykirsch:
Fixes #20 

Add a public `addCollector` method to the Agent.


README.md
```diff
@@ -19,6 +19,39 @@ Add the ServiceProvider class to the providers array in `config/app.php`:
...
+```php
+public function middleware()
+{
+    return [
+        app(\AG\ElasticApmLaravel\Jobs\Middleware\RecordTransaction::class),
+    ];
+}
+```
+
+Finally, you can add your own collector to the agent!
```
arkaitzgarro, line 50: Maybe we should also mention that the instance must implement `AG\ElasticApmLaravel\Collectors\Interfaces\DataCollectorInterface`? And I’m also thinking if we should explain how to measure times within the collector, because it can be tricky.

src/Agent.php
```diff
@@ -39,16 +39,18 @@ public function registerCollectors(Application $app): void
...
         // Http request collector
+        $this->addCollector(new HttpRequestCollector($app, $this->request_start_time));
+    }
+
+    public function addCollector(DataCollectorInterface $collector)
+    {
         $this->collectors->put(
-            HttpRequestCollector::getName(),
-            new HttpRequestCollector($app, $this->request_start_time)
+            $collector->getName(),
```
arkaitzgarro, line 56: Does this work, taking into account that the method is defined as static?

cykirsch: I had the same question! So I double-checked in tinker before I pushed it up. Works both ways.
```
>>> $x = new AG\ElasticApmLaravel\Collectors\DBQueryCollector(app(), 1.0);
=> AG\ElasticApmLaravel\Collectors\DBQueryCollector {#3879}
>>> echo $x->getName();
query-collector⏎
>>> echo $x::getName();
query-collector⏎
>>> exit
```

README.md
cykirsch: I can do that. I need a bit more direction on the time explanation, just not sure what specifically you'd like explained.

src/Agent.php
arkaitzgarro: In that case, maybe we should change the interface, it doesn't make sense anymore to have a `static` method definition. But since it is a breaking change, let's create a new interface following Laravel's folder and naming structure: `AG\ElasticApmLaravel\Contracts\DataCollector`

cykirsch:
I’d hesitate on that, but really only because the transaction stack changes
I played with also would have made a breaking change there. Not that those
specific changes are coming, just a caution on changing interfaces too
quickly.

Since it’s not causing a problem as it is—if the suggestion is to keep the
method but remove static, maybe we could leave it as is but document that
there is a planned change the next time the interface changes otherwise?
But I guess then people will use it now and have a less seamless upgrade at
that time.

So I’ll do whatever you think is best here but wanted to give my input
first.

arkaitzgarro:
While I agree with your reasoning, I think that in that specific case, it is ok to change the interface. It was only used internally, and never exposed in any public method, there was no reasoning on implementing this interface outside the internal code.

Now that we will expect arguments implementing this interface, I would rather prefer to have it in a final shape.

README.md
arkaitzgarro: I was writing some docs to explain how to implement a custom collector when I realized that start time is relative to the start of the transactions, and we don't make this value available. If a correct start time is not provided, the spans will not be rendered correctly.

We could solve this problem by doing the following:

1. Create a public method in `Agent` class that makes the `$request_start_time` available.
2. Create an abstract class (or even make `TimelineDataCollector` abstract) and ask developers to extend that class that already provides functionality to collect measures. This class will get `$request_start_time` from the Agent, instead of expecting the value as a constructor argument.
3. Add a simple example of how to create a custom collector.

cykirsch: I agree on all fronts, I'm working on 1 & 2 right now but also getting the abstract parent to use DI more effectively. Let me know how you like that.

cykirsch: That's done. As for 3, not sure how to give an example any simpler than the SpanCollector :joy: Tangentially, I did poke around to find a decent list of [events built-in to Laravel](https://laracasts.com/discuss/channels/general-discussion/where-can-i-get-a-complete-list-of-laravel-5-events?page=1#reply=503332) if you're looking for more ways to expand the package 🤷‍♂.

Anyway, the simplest explanation to me is linking to SpanCollector with a brief description of how it uses startMeasure/stopMeasure and to DBQueryCollector with a brief description of how it uses addMeasure directly.

src/Agent.php
```diff
@@ -92,4 +91,9 @@ public function sendTransaction(string $transaction_name): bool
 
         return parent::send();
     }
+
+    public function getRequestStartTime()
```
arkaitzgarro, line 95: Return type is missing.

src/Collectors/TimelineDataCollector.php
```diff
@@ -2,25 +2,49 @@
...
         $this->started_measures = new Collection();
         $this->measures = new Collection();
-        $this->request_start_time = $request_start_time;
+
+        $this->request_start_time = $this->agent->getRequestStartTime();
+
+        $this->registerEventListeners();
+    }
+
+    public function getName(): string
```
arkaitzgarro, line 45: We might want to remove this implementation, to _force_ child classes to provide a name for the custom collector.

arkaitzgarro (approved):
I'm going to merge and write some tests for collectors in a different PR.

PR #25, opened 2020-02-13
Make captureThrowable available everywhere

cykirsch:
Add pass-through captureThrowable method to the ApmCollector service/facade


src/Services/ApmCollectorService.php
```diff
@@ -75,4 +77,13 @@ public function addCollector(string $collector_class): void
...
         );
     }
+
+    public function captureThrowable(Throwable $thrown, array $context = [], ?Transaction $parent = null)
+    {
+        if ($this->is_agent_disabled) {
+            return;
+        }
+
+        $this->app->make(Agent::class)->captureThrowable($thrown, $context, $parent);
```
arkaitzgarro, line 87: Not sure how Laravel's service container works, but I hope this only creates one instance of the Agent that will be reused on every call to `captureThrowable`.

cykirsch: Correct, there is only one Agent instance because the AG Agent is registered as a singleton in the ServiceProvider.

arkaitzgarro (approved):

PR #26, opened 2020-02-15
chore: rename timeline collector

arkaitzgarro:
While writing unit tests for the `TimelineCollector` class, I realized about the following:

- The class is dedicated to listening to events, not to collect _timeline_ events. For that reason, I renamed the class to `EventDataCollector`. I also added a new method to the interface to reflect that behavior.
- It wasn't clear from the docs when to register the new collector. I tried to make it easier for our users, and I exposed a method in our Facade to add custom collectors: `ApmCollector::addCollector(MyCollector::class);`. The README includes an example to create and register an event collector.

@cykirsch thoughts?

@countless-integers I removed namespaces in unit tests, not sure how useful it is for classes that are never imported. I moved all tests into a `unit` folder, and I had to update all namespaces to make it work again...

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Add the ServiceProvider class to the providers array in `config/app.php`:
 From here, we will take care of everything based on your configuration. The agent and the middleware will be registered, and transactions will be sent to Elastic.
 
 ## Collectors
+
 The default collectors typically listen on events to measure portions of the request such as framework loading, database queries, or jobs.
 
 The SpanCollector in particular allows you to measure any section of your own code via the `ApmCollector` Facade:
@@ -47,11 +48,59 @@ public function middleware()
 }
 ```
 
-Finally, you can add your own collector to the agent!
+### Add a collector for other events
+
+You can add extra collector(s) to listen to your own application events or Laravel events like `Illuminate\Mail\Events\MessageSending` for example. We created a base collector that already includes functionality to measure events, that you can extend from:
+
+```php
+// app/Collectors/MailMessageCollector.php
+
+namespace YourApp\Collectors;
+
+use AG\ElasticApmLaravel\Contracts\DataCollector;
+use AG\ElasticApmLaravel\Collectors\EventDataCollector;
+
+use Illuminate\Mail\Events\MessageSending;
+use Illuminate\Mail\Events\MessageSent;
+
+class MailMessageCollector extends EventDataCollector implements DataCollector
+{
+    public function getName(): string
+    {
+        return 'mail-message-collector';
+    }
+
+    protected function registerEventListeners(): void
+    {
+        $this->app->events->listen(MessageSending::class, function (\Swift_Message $message) {
+            $this->startMeasure(
+                'mail #' . $message->getId(),
+                'mail.delivery',
+            );
+        });
+
+        $this->app->events->listen(StopMeasuring::class, function (\Swift_Message $message) {
+            $this->stopMeasure('mail #' . $message->getId());
+        });
+    }
+}
 
 ```
-$agent = app(\AG\ElasticApmLaravel\Agent::class);
-$agent->addCollector(new MyCollector());
+
+Don't forget to register your collector when the application starts:
+
+```php
+// app/Providers/AppServiceProvider.php
+
+use AG\ElasticApmLaravel\Facades\ApmCollector;
+
+use YourApp\Collectors\MailMessageCollector;
+
+public function boot()
+{
+    // ...
+    ApmCollector::addCollector(MailMessageCollector::class);
+}
 ```
 
 ## Agent configuration
--- a/src/Collectors/EventDataCollector.php
+++ b/src/Collectors/EventDataCollector.php
@@ -4,22 +4,15 @@ namespace AG\ElasticApmLaravel\Collectors;
 
 use AG\ElasticApmLaravel\Agent;
 use AG\ElasticApmLaravel\Contracts\DataCollector;
-use Illuminate\Foundation\Application;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Log;
 
 /**
- * Collects info about the request duration as well as providing
- * a way to log duration of any operations.
+ * Abstract class that provides base functionality to measure
+ * events dispatched by the framnewrok or your application code.
  */
-abstract class TimelineDataCollector implements DataCollector
+abstract class EventDataCollector implements DataCollector
 {
-    /** @var Application */
-    protected $app;
-
-    /** @var Agent */
-    protected $agent;
-
     /** @var Collection */
     protected $started_measures;
 
@@ -29,22 +22,14 @@ abstract class TimelineDataCollector implements DataCollector
     /** @var float */
     protected $request_start_time;
 
-    public function __construct(Application $app, Agent $agent)
+    public function __construct(Agent $agent)
     {
-        $this->app = $app;
-        $this->agent = $agent;
-
         $this->started_measures = new Collection();
         $this->measures = new Collection();
 
-        $this->request_start_time = $this->agent->getRequestStartTime();
+        $this->request_start_time = $agent->getRequestStartTime();
 
         $this->registerEventListeners();
-    }
-
-    public function getName(): string
-    {
-        return 'timeline';
     }
 
     /**
@@ -131,10 +116,6 @@ abstract class TimelineDataCollector implements DataCollector
         });
 
         return $this->measures;
-    }
-
-    protected function registerEventListeners(): void
-    {
     }
 
     private function toMilliseconds(float $time): float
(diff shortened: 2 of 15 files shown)


tests/unit/Services/ApmCollectorServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+
+                return true;
+            }));
+
+        $this->collectorService->startMeasure(self::SPAN_NAME);
+    }
+
+    public function testStartMeasureEventWithValues()
+    {
+        $this->eventsMock->shouldReceive('dispatch')
```
cykirsch, line 60: This accomplishes the same thing but another good way to assert events are dispatched is to register a listener inside the test method and set a value to check after the SUT runs.

README.md
```diff
@@ -47,11 +48,59 @@ public function middleware()
...
+Don't forget to register your collector when the application starts:
+
+```php
+// app/Providers/AppServiceProvider.php
+
+use AG\ElasticApmLaravel\Facades\ApmCollector;
+
+use YourApp\Collectors\MailMessageCollector;
+
+public function boot()
```
cykirsch, line 99: I like this recommendation in boot instead of register and think we should put the default collectors under boot as well.

src/Collectors/EventDataCollector.php
```diff
@@ -4,22 +4,15 @@
...
 use Illuminate\Support\Facades\Log;
 
 /**
- * Collects info about the request duration as well as providing
- * a way to log duration of any operations.
+ * Abstract class that provides base functionality to measure
+ * events dispatched by the framnewrok or your application code.
  */
-abstract class TimelineDataCollector implements DataCollector
+abstract class EventDataCollector implements DataCollector
```
cykirsch, line 15: I like the rename.

src/Collectors/EventDataCollector.php
```diff
@@ -29,24 +22,16 @@ abstract class TimelineDataCollector implements DataCollector
     /** @var float */
     protected $request_start_time;
 
-    public function __construct(Application $app, Agent $agent)
```
cykirsch: The child classes use $app but no longer have it.

cykirsch (changes requested):
Looks good, the only issue I can see is the missing $app in collector implementations.

src/Collectors/EventDataCollector.php
arkaitzgarro: You are absolutely right. Adding it right now.

README.md
arkaitzgarro: If I understand correctly `register/boot` methods, we should register the agent and facade in the `register` method, and the middleware and collectors in the `boot` method. Am I correct?

tests/unit/Services/ApmCollectorServiceTest.php
arkaitzgarro: I'm not sure what you mean. Do you have a concrete example?

README.md
cykirsch: Yep anything that describes how an object should be instantiated goes in register, and the usages of those objects can be in boot.

tests/unit/Services/ApmCollectorServiceTest.php
cykirsch: I’m traveling and on my phone so this is real rough...

$dispatched = false;
App()->listen(MyEvent::class, function() {
  $dispatched = true;
});

$collector->startMeasure(...);
$this->assertTrue($dispatched);

But it reaches the same result so it’s just a matter of preference.

cykirsch: I’m traveling and on my phone so this is real rough...

$dispatched = false;
App()->listen(MyEvent::class, function() {
  $dispatched = true;
});

$collector->startMeasure(...);
$this->assertTrue($dispatched);

But it reaches the same result so it’s just a matter of preference.

arkaitzgarro: I see. In that case, we are testing that not only the event was fired, but also if the data passed through the event is correct.

```php
$this->assertEquals(self::SPAN_NAME, $event->name);
$this->assertEquals('request', $event->type);
$this->assertNull($event->action);
$this->assertNull($event->label);
$this->assertNull($event->start_time);
```

tests/unit/Collectors/CustomCollectorTest.php
```diff
@@ -1,26 +1,47 @@
...
      */
     private $collector;
 
     protected function _before()
     {
-        $this->collector = new TimelineDataCollector(0);
+        $appMock = Mockery::mock(Application::class);
+        $agentMock = Mockery::mock(Agent::class);
+
+        $agentMock->shouldReceive('getRequestStartTime')->andReturn(1000.0);
```
countless-integers: Missing `->once()` or how ever many times you expect this to be called.

tests/unit/Collectors/CustomCollectorTest.php
```diff
@@ -1,26 +1,47 @@
...
-use AG\ElasticApmLaravel\Collectors\TimelineDataCollector;
+use AG\ElasticApmLaravel\Agent;
+use AG\ElasticApmLaravel\Collectors\EventDataCollector;
 use Codeception\Test\Unit;
 use DMS\PHPUnitExtensions\ArraySubset\Assert;
+use Illuminate\Foundation\Application;
 use Illuminate\Support\Facades\Log;
 
-class TimelineDataCollectorTest extends Unit
+class CustomCollector extends EventDataCollector
```
countless-integers, line 10: could also use an [anonymous class](https://www.php.net/manual/en/language.oop5.anonymous.php) instead.

tests/unit/Collectors/CustomCollectorTest.php
```diff
@@ -47,7 +68,7 @@ public function testStartMeasure()
         $this->assertEquals(1, $this->collector->collect()->count());
         Assert::assertArraySubset([
             'label' => 'GET /endpoint',
-            'start' => 1000000.0,
+            'start' => 0.0,
```
countless-integers, line 74: that's quite a change...

tests/unit/Services/ApmCollectorServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+        );
+    }
+
+    public function testAddCollector()
+    {
+        $agentMock = Mockery::mock(Agent::class);
+        $collectorMock = Mockery::mock(DataCollector::class);
+
+        $agentMock->shouldReceive('addCollector')
+            ->with($collectorMock);
```
countless-integers, line 119:
```suggestion
            ->once()
            ->with($collectorMock);
```

tests/unit/Services/ApmCollectorServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $this->appMock->shouldReceive('make')
+            ->with(Agent::class)
```
countless-integers, line 123:
```suggestion
            ->once()
            ->with(Agent::class)
```

tests/unit/Services/ApmCollectorServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+            ->andReturn($agentMock);
+
+        $this->appMock->shouldReceive('make')
+            ->with('App\Collectors\MyCollector')
```
countless-integers, line 128:
```suggestion
            ->once()
            ->with('App\Collectors\MyCollector')
```

tests/unit/Services/ApmCollectorServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+            ->andReturn($collectorMock);
+
+        $this->collectorService->addCollector('App\Collectors\MyCollector');
```
countless-integers, line 131: Would this also work with a class const? `MyCollector::class`?

countless-integers (changes requested):
More like a refactor than a chore.

tests/unit/Collectors/CustomCollectorTest.php
arkaitzgarro: I thinks it's cleaner to have a separate class definition. Otherwise it will defined inside the `_before` method. feels a bit messy.

tests/unit/Collectors/CustomCollectorTest.php
arkaitzgarro: `request_start_time` is now `1000`, and the span starts at the same time. That's why it went down to `0.0`.

tests/unit/Services/ApmCollectorServiceTest.php
arkaitzgarro: Yes :)

tests/unit/Collectors/CustomCollectorTest.php
countless-integers: 🤷‍♂ fair enough.

tests/unit/Services/ApmCollectorServiceTest.php
countless-integers: 🤞

countless-integers (approved):

PR #29, opened 2020-02-20
chore: add unit tests for request Middleware

arkaitzgarro:
Worth to mention that I'm injecting the configuration instead of using Laravel's global functions. I'm not sure if there is an easy way to mock global function tbh.

### DIFF
--- a/src/Middleware/RecordTransaction.php
+++ b/src/Middleware/RecordTransaction.php
@@ -4,9 +4,9 @@ namespace AG\ElasticApmLaravel\Middleware;
 
 use AG\ElasticApmLaravel\Agent;
 use Closure;
-use Illuminate\Config\Repository as Config;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Route;
+use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Log;
 use PhilKra\Events\Transaction;
 use Symfony\Component\HttpFoundation\Response;
@@ -26,12 +26,10 @@ use Throwable;
 class RecordTransaction
 {
     protected $agent;
-    protected $config;
 
-    public function __construct(Agent $agent, Config $config)
+    public function __construct(Agent $agent)
     {
         $this->agent = $agent;
-        $this->config = $config;
     }
 
     /**
@@ -47,7 +45,7 @@ class RecordTransaction
         // Execute the application logic
         $response = $next($request);
 
-        if ($this->config->get('elastic-apm-laravel.transactions.useRouteUri')) {
+        if (Config::get('elastic-apm-laravel.transactions.useRouteUri')) {
             $transaction->setTransactionName($this->getRouteUriTransactionName($request));
         }
 
--- a/src/Middleware/RecordTransaction.php
+++ b/src/Middleware/RecordTransaction.php
@@ -4,6 +4,7 @@ namespace AG\ElasticApmLaravel\Middleware;
 
 use AG\ElasticApmLaravel\Agent;
 use Closure;
+use Illuminate\Config\Repository as Config;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Route;
 use Illuminate\Support\Facades\Log;
@@ -25,10 +26,12 @@ use Throwable;
 class RecordTransaction
 {
     protected $agent;
+    protected $config;
 
-    public function __construct(Agent $agent)
+    public function __construct(Agent $agent, Config $config)
     {
         $this->agent = $agent;
+        $this->config = $config;
     }
 
     /**
@@ -44,7 +47,7 @@ class RecordTransaction
         // Execute the application logic
         $response = $next($request);
 
-        if (config('elastic-apm-laravel.transactions.useRouteUri')) {
+        if ($this->config->get('elastic-apm-laravel.transactions.useRouteUri')) {
             $transaction->setTransactionName($this->getRouteUriTransactionName($request));
         }
 
--- a/src/Middleware/RecordTransaction.php
+++ b/src/Middleware/RecordTransaction.php
@@ -4,10 +4,10 @@ namespace AG\ElasticApmLaravel\Middleware;
 
 use AG\ElasticApmLaravel\Agent;
 use Closure;
+use Illuminate\Config\Repository as Config;
 use Illuminate\Http\Request;
+use Illuminate\Log\Logger;
 use Illuminate\Routing\Route;
-use Illuminate\Support\Facades\Config;
-use Illuminate\Support\Facades\Log;
 use PhilKra\Events\Transaction;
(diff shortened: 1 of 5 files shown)


cykirsch:
> Worth to mention that I'm injecting the configuration instead of using Laravel's global functions. I'm not sure if there is an easy way to mock global function tbh.

Yes you can. It uses the Facade underneath, which have a mocking feature. https://laravel.com/docs/6.x/facades#facades-vs-helper-functions

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -0,0 +1,145 @@
...
+        $this->agent->shouldReceive('collectEvents')
+            ->once()
+            ->with('GET /ping');
+
+        $this->middleware->terminate($this->request);
+    }
+
+    public function testTransactionTerminateError()
+    {
+        $this->agent->shouldReceive('stopTransaction')
```
countless-integers, line 173:
```suggestion
        $this->agent->shouldReceive('stopTransaction')
            ->once()
```

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -0,0 +1,145 @@
...
+        $this->assertEquals('HTTP', $this->transaction->getMetaType());
+    }
+
+    public function testTransactionContext()
+    {
+        $this->agent->shouldReceive('startTransaction')
+            ->andReturn($this->transaction);
+
+        $this->response->shouldReceive('getStatusCode')
+            ->andReturn(200);
```
countless-integers, line 115:
```suggestion
            ->once()
            ->andReturn(200);
```

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -0,0 +1,145 @@
...
+        });
+
+        $this->assertEquals(200, $this->transaction->getMetaResult());
+        $this->assertEquals('HTTP', $this->transaction->getMetaType());
+    }
+
+    public function testTransactionContext()
+    {
+        $this->agent->shouldReceive('startTransaction')
+            ->andReturn($this->transaction);
```
countless-integers, line 111:
```suggestion
            ->once()
            ->andReturn($this->transaction);
```

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -0,0 +1,145 @@
...
+        });
+    }
+
+    public function testTransactionMetadata()
+    {
+        $this->agent->shouldReceive('startTransaction')
+            ->andReturn($this->transaction);
+
+        $this->response->shouldReceive('getStatusCode')
+            ->andReturn(200);
```
countless-integers, line 95:
```suggestion
            ->once()
            ->andReturn(200);
```

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -0,0 +1,145 @@
...
+            ->andReturn($this->transaction);
+
+        $this->middleware->handle($this->request, function () {
+            return $this->response;
+        });
+    }
+
+    public function testTransactionMetadata()
+    {
+        $this->agent->shouldReceive('startTransaction')
```
countless-integers, line 89:
```suggestion
        $this->agent->shouldReceive('startTransaction')
            ->once()
```

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -0,0 +1,145 @@
...
+    {
+        $this->agent = Mockery::mock(Agent::class);
+        $this->config = Mockery::mock(Config::class);
+        $this->transaction = Mockery::mock(Transaction::class)->makePartial();
+        $this->request = Request::create('/ping', 'GET');
+        $this->response = Mockery::mock(Response::class)->makePartial();
+        $this->response->headers = new ResponseHeaderBag();
+
+        $this->config->shouldReceive('get')
+            ->with('elastic-apm-laravel.transactions.useRouteUri')
```
countless-integers, line 55:
```suggestion
            ->once()
            ->with('elastic-apm-laravel.transactions.useRouteUri')
```

countless-integers (changes requested):
Just my favourite `once()` topic. What @cykirsch mention stands as well ([docs](https://laravel.com/docs/5.5/mocking#mocking-facades)). We could just use facedes instead of those global handlers from this point on to have some extra testing options.

src/Middleware/RecordTransaction.php
```diff
@@ -44,7 +45,7 @@ public function handle(Request $request, Closure $next)
         // Execute the application logic
         $response = $next($request);
 
-        if (config('elastic-apm-laravel.transactions.useRouteUri')) {
+        if (Config::get('elastic-apm-laravel.transactions.useRouteUri')) {
```
arkaitzgarro: I'm having some trouble to make `Config::get` return `true` in one specific spec. I'm using `_before` to always return `false`:

```php
protected function _before()
{
    Config::shouldReceive('get')
        ->with('elastic-apm-laravel.transactions.useRouteUri')
        ->andReturn(false);
    ...
}
```

But in one specific spec I want to return `true`:

```php
public function testUriTransactionName()
{
    Config::shouldReceive('get')
        ->with('elastic-apm-laravel.transactions.useRouteUri')
        ->andReturn(true);
    ...
    $this->middleware->handle($this->request, function () {
        return $this->response;
    });
}
```

@cykirsch any idea how to achieve that with Facades?

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -0,0 +1,147 @@
...
+            'id' => null,
+            'ip' => '127.0.0.1',
+            'user-agent' => 'Symfony/3.X',
+        ], $context['user']);
+    }
+
+    public function testTransactionTerminate()
+    {
+        // Don't expect Config::get to be called
+        Config::mockery_teardown();
```
countless-integers: wow, is this a work-around for changing mockery expectation? I'm not sure I'd not prefer to just copy-paste the expectation from `_before` to all methods that need it.

src/Middleware/RecordTransaction.php
cykirsch: I haven't actually tried to do that pattern, and was surprised to see it works for you in some of the other tests I've looked at in the package. So no, I don't know how to achieve it with facades.

I'd side with @countless-integers https://github.com/arkaitzgarro/elastic-apm-laravel/pull/29/files#r383183831. Prevents any need for workarounds in both of these cases, and it's very explicit. If the duplication is bothersome it can go in a helper method so it's at least a shorter call to make.

arkaitzgarro:
@countless-integers @cykirsch I gave up with Facades and I'm using good old DI for configuration and logging. That makes it easier to create a clean and shiny instance of configuration for every spec.

PR #30, opened 2020-02-20
Fixes for framework events and job spans

cykirsch:
Moving the `addCollector` call to the `boot` method caused the framework events to start being counted wrong because those listeners need to be registered before the Application booting and booted events, and the boot method is too late.
In addition, in my prior job testing, I was only using the `sync` driver, which is not very real-world behavior. With queued drivers like database, they process in a worker, so the agent doesn't destruct after one job and therefore events need to be sent manually after each of these jobs.

### DIFF
--- a/src/Collectors/JobCollector.php
+++ b/src/Collectors/JobCollector.php
@@ -3,9 +3,12 @@
 namespace AG\ElasticApmLaravel\Collectors;
 
 use AG\ElasticApmLaravel\Contracts\DataCollector;
+use Illuminate\Contracts\Queue\Job;
+use Illuminate\Queue\Events\JobExceptionOccurred;
 use Illuminate\Queue\Events\JobFailed;
 use Illuminate\Queue\Events\JobProcessed;
 use Illuminate\Queue\Events\JobProcessing;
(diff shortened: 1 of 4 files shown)


src/Collectors/JobCollector.php
```diff
@@ -68,12 +79,23 @@ protected function setTransactionResult(string $transaction_name, int $result):
...
+    protected function send(Job $job): void
     {
         try {
-            // Stop the transaction and measure the time
-            $this->agent->stopTransaction($transaction_name);
-            $this->agent->collectEvents($transaction_name);
+            if (! ($job instanceof SyncJob)) {
+                // When using a queued driver, send/flush transaction to make room for the next job in the queue
+                // Otherwise just send when the agent destructs
+                $this->agent->send();
```
cykirsch, line 88: I really wanted to do this in the `Looping` event. Unfortunately that runs before picking up a job, so no transaction has been started (it fails if you send() with no transaction). And we can't check to see if there is a transaction because the transaction store is private, and the Looping event doesn't have a job in it so we can't check by name.

arkaitzgarro (approved):

PR #31, opened 2020-02-28
chore: fix unit tests on master

arkaitzgarro:
Somehow unit tests passed in the PR, but not on master.


countless-integers (approved):
ok, so basically some of the tests didn't expect that method call. Weird that it would pass then...

arkaitzgarro:
> ok, so basically some of the tests didn't expect that method call. Weird that it would pass then...

The only reason I can think of, is that we introduced the Mockery expectations after the PR was created...

PR #32, opened 2020-03-03
fix: use Log facade on transaction middleware

arkaitzgarro:
We were injecting a class that is not available on Laravel 5.5.


countless-integers (approved):
Sad, but true...

PR #33, opened 2020-03-04
fix: correct times for async jobs

arkaitzgarro:
For jobs executed by queue workers, we are considering the `request_start_time` variable as the start of the request time, but this is not valid for processes that run in the background, like queue workers. Transaction times will grow in time:

Screenshot 2020-03-04 at 15 09 24

For transactions executed via cli, take the job starting time as transaction starting time, and don't measure Laravel boot times.

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -40,7 +40,11 @@ class Agent extends PhilKraAgent
     public function registerInitCollectors(): void
     {
         // Laravel init collector
-        $this->addCollector(app(FrameworkCollector::class));
+        if ('cli' !== php_sapi_name()) {
+            // For cli executions, like queue workers, the application
+            // only starts once. It doesn't really make sense to measure it.
+            $this->addCollector(app(FrameworkCollector::class));
+        }
     }
 
     public function registerCollectors(): void
--- a/tests/unit/Collectors/JobCollectorTest.php
+++ b/tests/unit/Collectors/JobCollectorTest.php
@@ -72,7 +72,13 @@ class JobCollectorTest extends Unit
         $this->agentMock
             ->shouldReceive('startTransaction')
             ->once()
-            ->with(self::JOB_NAME, [], self::REQUEST_START_TIME)
+            ->withArgs(function ($job_name, $context, $start_time) {
+                $this->assertEquals(self::JOB_NAME, $job_name);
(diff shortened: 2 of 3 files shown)


src/Agent.php
```diff
@@ -40,7 +40,11 @@ public function __construct(array $config, float $request_start_time)
     public function registerInitCollectors(): void
     {
         // Laravel init collector
-        $this->addCollector(app(FrameworkCollector::class));
+        if ('cli' !== php_sapi_name()) {
```
countless-integers, line 43: Thought: if the collector is not initialised for cli requests should it still be called a "FrameworkCollector"?

tests/unit/Collectors/JobCollectorTest.php
```diff
@@ -72,7 +72,13 @@ public function testJobProcessingListener()
         $this->agentMock
             ->shouldReceive('startTransaction')
             ->once()
-            ->with(self::JOB_NAME, [], self::REQUEST_START_TIME)
+            ->withArgs(function ($job_name, $context, $start_time) {
+                $this->assertEquals(self::JOB_NAME, $job_name);
+                $this->assertEquals([], $context);
+                $this->assertNotNull($start_time);
```
countless-integers, line 78: interesting testing problem... would have to spoof `php_sapi_name` or abstract it that it uses the correct time for a specific env.

countless-integers (approved):
Got some feedback, but it'd be nice to get it working first.

src/Agent.php
arkaitzgarro: it is still a FrameworkCollector, but since the framework only loads once for workers, not sure about the value that it adds. I would like to measure artisan commands as well, maybe we can revisit the approach at that point.

tests/unit/Collectors/JobCollectorTest.php
arkaitzgarro: Do you like this? https://marcelog.github.io/articles/php_mock_global_functions_for_unit_tests_with_phpunit.html

cykirsch: FWIW I had the same thoughts when using the method and when I looked around the general recommendation was to wrap it in your own method and call/mock the wrapper instead.

PR #35, opened 2020-03-06
Configurable filtering of transactions by name

cykirsch:
My logs are full of the error `The transaction "OPTIONS /my/api/path" is not registered.` I have not been able to figure out where the `getTransaction()` is initiated in this case. It likely has something to do with the barryvdh/laravel-cors package, which is worth supporting since that behavior is baked into Laravel 7.

I'm completely open to suggestions on a better solution, but this works well to fix the behavior I'm seeing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ The following environment variables are supported in the default configuration:
 |APM_SERVERURL      | URL to the APM intake service. |
 |APM_SECRETTOKEN    | Secret token, if required. |
 |APM_USEROUTEURI    | `true` or `false` defaults to `true`. The default behavior is to record the URL as defined in your routes configuration. Set to `false` to record the requested URL, but keep in mind that this can result in excessive unique entries in APM. |
-|APM_IGNORE_PATTERNS| Ignore specific routes by transaction name. Should be a regular expression, and will match multiple patterns via pipe `|` in the regex. Example: `"/\/health-check|^OPTIONS /"` |
+|APM_IGNORE_PATTERNS| Ignore specific routes by transaction name. Should be a regular expression, and will match multiple patterns via pipe `\|` in the regex. Example: `"/\/health-check\|^OPTIONS /"` |
 |APM_QUERYLOG       | `true` or `false` defaults to 'true'. Set to `false` to completely disable query logging, or to `auto` if you would like to use the threshold feature. |
 |APM_THRESHOLD      | Query threshold in milliseconds, defaults to `200`. If a query takes longer then 200ms, we enable the query log. Make sure you set `APM_QUERYLOG=auto`. |
 |APM_BACKTRACEDEPTH | Defaults to `25`. Depth of backtrace in query span. |
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ The following environment variables are supported in the default configuration:
 |APM_SERVERURL      | URL to the APM intake service. |
 |APM_SECRETTOKEN    | Secret token, if required. |
 |APM_USEROUTEURI    | `true` or `false` defaults to `true`. The default behavior is to record the URL as defined in your routes configuration. Set to `false` to record the requested URL, but keep in mind that this can result in excessive unique entries in APM. |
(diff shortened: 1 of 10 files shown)


arkaitzgarro (commented):
This solution is a bit drastic, other users could be interested in OPTIONS requests. 

Ideally, we offer the possibility to filter out transactions based on regular expressions, like the Python agent does: https://github.com/elastic/apm-agent-python/blob/5.x/docs/configuration.asciidoc#transactions_ignore_patterns

countless-integers:
I like the filter idea. Or at least make it configurable -- little work necessary.

cykirsch:
@arkaitzgarro @countless-integers Thanks for the feedback! The link to the python agent was especially helpful. As of now I've still only included support for filtering on the HTTP transactions. Let me know how you like the changes.

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -47,12 +47,15 @@ protected function _before()
         $this->response->headers = new ResponseHeaderBag();
     }
 
-    protected function createMiddlewareInstance(bool $use_route_uri): void
+    protected function createMiddlewareInstance(bool $use_route_uri, ?string $ignore_patterns): void
```
countless-integers: Is this a deliberate choice over setting a default value of `''`? Default value would spare us from changing `createMiddlewareInstance` invocations.

tests/unit/Middleware/RecordTransactionTest.php
```diff
@@ -47,12 +47,15 @@ protected function _before()
...
     {
         $this->config = Mockery::mock(Config::class);
         $this->config->shouldReceive('get')
             ->with('elastic-apm-laravel.transactions.useRouteUri')
             ->andReturn($use_route_uri);
+        $this->config->shouldReceive('get')
+            ->with('elastic-apm-laravel.transactions.ignorePatterns')
```
countless-integers, line 57: missing my favourite Mockery method:

```suggestion
            ->with('elastic-apm-laravel.transactions.ignorePatterns')
            ->once()
```

countless-integers (changes requested):
I like the pattern solution 👍

arkaitzgarro:
> @arkaitzgarro @countless-integers Thanks for the feedback! The link to the python agent was especially helpful. As of now I've still only included support for filtering on the HTTP transactions. Let me know how you like the changes.

Thanks! It would be really nice to apply the same logic for jobs as well.

tests/unit/Middleware/RecordTransactionTest.php
cykirsch: Not deliberate. Fixed.

cykirsch:
@arkaitzgarro Job filtering is now complete.

README.md
```diff
@@ -118,6 +118,7 @@ The following environment variables are supported in the default configuration:
 |APM_SERVERURL      | URL to the APM intake service. |
 |APM_SECRETTOKEN    | Secret token, if required. |
 |APM_USEROUTEURI    | `true` or `false` defaults to `true`. The default behavior is to record the URL as defined in your routes configuration. Set to `false` to record the requested URL, but keep in mind that this can result in excessive unique entries in APM. |
+|APM_IGNORE_PATTERNS| Ignore specific routes or jobs by transaction name. Should be a regular expression, and will match multiple patterns via pipe `\|` in the regex. Note that 4 backslashes should be used to match a single backslash. Example: `"/\/health-check\|^OPTIONS \|Foo\\\\Bar\\\\Job/"` |
```
countless-integers, line 121: btw. considering that this pattern will likely have slashes in it, it might also be a nice option to use a different pattern delimiter. E.g. you could do `@/foo/bar@` rather than `/\/foo\/bar/`. No change required, since this is possible with the current setup.

countless-integers (approved):

arkaitzgarro (approved):
It would be amazing if this could be done by `philkra/elastic-apm-php-agent` so we don't have to do anything. @cykirsch could you perhaps open a PR and apply the filter in the `send()` method?

PR #51, opened 2020-06-29
Switch to Events\Dispatcher interface to support alternate / mock event dispatchers

hubie:
I ran into an issue with running an integration test that verified events using Laravel's [Event::fake()](https://laravel.com/docs/7.x/mocking#event-fake).  Because `Event::fake()` replaces `Illuminate\Events\Dispatcher` with `\Illuminate\Support\Testing\Fakes\EventFake`, I was receiving the following error:
`1) mytest.php::test_a_thing
  TypeError: Argument 2 passed to AG\ElasticApmLaravel\Services\ApmCollectorService::__construct() must be an instance of Illuminate\Events\Dispatcher, instance of Illuminate\Support\Testing\Fakes\EventFake given`

By switching to the Contract, the CollectorService will be able to operate with both `Illuminate\Events\Dispatcher` (which Laravel binds by default) as well as the `EventFake`.

### DIFF
--- a/src/Services/ApmCollectorService.php
+++ b/src/Services/ApmCollectorService.php
@@ -6,7 +6,7 @@ use AG\ElasticApmLaravel\Agent;
 use AG\ElasticApmLaravel\Events\StartMeasuring;
 use AG\ElasticApmLaravel\Events\StopMeasuring;
 use Illuminate\Config\Repository as Config;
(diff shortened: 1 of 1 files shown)


cykirsch (approved):

arkaitzgarro (approved):
Great, thank you very much!

PR #60, opened 2020-07-27
fix: make sure transaction object exists before sending error events

countless-integers:
I've noticed that we're making 2 api requests with events to APM and the second one lacks transaction data (but the transaction name is set). This raises an exception during handling an exception, so chaos ensues. From what I understand from the [documentation](https://www.elastic.co/guide/en/apm/server/current/error-api.html), transaction data or id is required, so there's no reason to send requests if it's `null`.

### DIFF
--- a/src/Collectors/JobCollector.php
+++ b/src/Collectors/JobCollector.php
@@ -47,11 +47,11 @@ class JobCollector extends EventDataCollector implements DataCollector
             $transaction_name = $this->getTransactionName($event);
             if ($transaction_name) {
                 $transaction = $this->getTransaction($transaction_name);
-                $this->agent->captureThrowable($event->exception, [], $transaction);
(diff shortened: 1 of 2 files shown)


arkaitzgarro (approved):

countless-integers:
@cykirsch since you edited it last, do you maybe remember if there was some legit use-case for having this behaviour in the first place? Wouldn't one this fix to break something else 😉

cykirsch:
@countless-integers Nothing is coming to mind. This seems like the right thing to do.

What causes the second API request?

countless-integers:
I don't know tbh. The only other clue would be that there's no metadata included with it. Almost as if it was triggered by another event. Anyway, this can further corrected in another PR. Merging this will at least remove the errors for us.

PR #62, opened 2020-08-28
chore: prevent duplicated transactions

arkaitzgarro:
Somehow, some job transactions can be started within the same process with the same name, something not allowed by the underlying agent. We don't really know the reason, but it throws an exception and we want to prevent that.


countless-integers (approved):
heh, let's give that a try.

PR #63, opened 2020-08-28
feat: add metadata to Job events

arkaitzgarro:
Add Job metadata to APM via custom context:

- `job_id`: the job identifier.
- `attempts`: the number of times the job has been attempted.
- `max_tries`: the number of times to attempt a job.
- `connection_name`: the name of the connection the job belongs to.
- `queue_name`: name of the queue the job belongs to.

Important: Custom metadata adds non-indexed, custom contextual information to transactions and errors. Non-indexed means the data is not searchable or aggregatable in Elasticsearch, and you cannot build dashboards on top of the data. This also means you don’t have to worry about mapping explosions, as these fields are not added to the mapping.

https://www.elastic.co/guide/en/apm/get-started/current/metadata.html#custom-fields

### DIFF
--- a/src/Collectors/JobCollector.php
+++ b/src/Collectors/JobCollector.php
@@ -41,6 +41,7 @@ class JobCollector extends EventDataCollector implements DataCollector
 
             $this->startTransaction($transaction_name);
             $this->setTransactionType($transaction_name);
+            $this->addMetadata($transaction_name, $event->job);
         });
 
         $this->app->events->listen(JobProcessed::class, function (JobProcessed $event) {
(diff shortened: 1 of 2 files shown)


src/Collectors/JobCollector.php
```diff
@@ -105,6 +106,17 @@ protected function stopTransaction(string $transaction_name, int $result): void
         $this->agent->collectEvents($transaction_name);
     }
 
+    protected function addMetadata(string $transaction_name, Job $job): void
+    {
+        $this->agent->getTransaction($transaction_name)->setCustomContext([
```
countless-integers, line 111: ok, so this is a setter, but [it actually merges existing context with this](https://github.com/philkra/elastic-apm-php-agent/blob/773ef84e5f524d0f7dc0950e9f434255258e09f5/src/Events/EventBean.php#L215) (which I think is good in this case).

src/Collectors/JobCollector.php
```diff
@@ -41,6 +41,7 @@ public function registerEventListeners(): void
 
             $this->startTransaction($transaction_name);
             $this->setTransactionType($transaction_name);
+            $this->addMetadata($transaction_name, $event->job);
```
countless-integers, line 44: Next iteration something like this to allow users to set their own metadata via a config-defined callback:

```php
if (Config::get('elastic-apm-laravel.transactions.metadata_provider')) {
    $this->addMetadata($transaction_name, call_user_func(Config::get('elastic-apm-laravel.transactions.metadata_provider'), $event));
}
```

countless-integers (approved):

PR #66, opened 2020-09-06
Chore/migrate php agent

dstepe:
This uses the new release of the `nipwaayoni/elastic-apm-php-agent` package to address some issues with extending the `Agent` class. 

I had to make some changes to how the `AG\ElasticApmLaravel\Agent`, but I hope the impact is limited to the service provider. I don't expect anyone to be creating an agent themselves.

The ServiceProvider has been updated to us the provided `AgentBuilder` to construct the agent object. While this is a few more lines of code, the `Nipwaayoni\Agent` had become far too complicated to create in a flexible way. The `Nipwaayoni\AgentBuilder` enables flexibility while hiding the complexity of the creation.

While implementing this, I did find a Laravel register/boot order issue which I would like to address in a separate PR before this is all merged for a release.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -6,7 +6,11 @@ use AG\ElasticApmLaravel\Contracts\VersionResolver;
 use AG\ElasticApmLaravel\Middleware\RecordTransaction;
 use AG\ElasticApmLaravel\Services\ApmCollectorService;
 use Illuminate\Contracts\Http\Kernel;
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Illuminate\Support\ServiceProvider as BaseServiceProvider;
+use Nipwaayoni\AgentBuilder;
+use Nipwaayoni\Config;
 
 class ServiceProvider extends BaseServiceProvider
 {
@@ -64,7 +68,28 @@ class ServiceProvider extends BaseServiceProvider
         $this->app->singleton(Agent::class, function () {
             $start_time = $this->app['request']->server('REQUEST_TIME_FLOAT') ?? microtime(true);
 
-            return new Agent($this->getAgentConfig(), $start_time);
+            /** @var AgentBuilder $builder */
+            $builder = $this->app->make(AgentBuilder::class);
+
+            $builder->withAgentClass(Agent::class);
+            $builder->withConfig(new Config($this->getAgentConfig()));
+
+            $builder->withEnvData(config('elastic-apm-laravel.env.env'));
+
+            /** @var Agent $agent */
+            $agent = $builder->build();
+
+            $agent->setRequestStartTime($start_time);
+
+            return $agent;
+        });
+
+        // Register a callback on terminating to send the events
+        $this->app->terminating(function (Request $request, Response $response) {
+            /** @var Agent $agent */
+            $agent = $this->app->make(Agent::class);
+
+            $agent->send();
         });
     }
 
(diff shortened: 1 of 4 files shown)


arkaitzgarro:
Hey @dstepe, thanks for the help, this is great. I'm going to merge it into my branch and test it on our application, I will let you know 👍

dstepe:
Sounds good. Do hold on merging anything to your master branch. I'm close to finishing another set of work for the switch, and that also involves rethinking how the Agent class extension is done. Someone on the agent project pointed out what should be a better approach.

src/ServiceProvider.php
```diff
@@ -64,7 +68,28 @@ protected function registerAgent(): void
...
+            /** @var Agent $agent */
+            $agent = $builder->build();
+
+            $agent->setRequestStartTime($start_time);
+
+            return $agent;
+        });
+
+        // Register a callback on terminating to send the events
+        $this->app->terminating(function (Request $request, Response $response) {
```
arkaitzgarro, line 88: Just to double-check, this will be executed after the middleware [terminate](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Middleware/RecordTransaction.php#L88) method right?

dstepe: Yes. It does not seem to be well documented, but that is what others have concluded:

https://dev.to/webong/execute-an-action-after-laravel-returns-response-4pjc

I prefer not to rely on the middleware `terminate` because the documentation implies it is only call when using FastCGI:

> If you define a terminate method on your middleware and your web server is using FastCGI, the terminate method will automatically be called after the response is sent to the browser

Practical experience indicates this works, but I have not proven the accuracy of the collected data. Additional validation would be great.

The "automatic" sending of data was removed from the `Agent` in the new package. Using the `Agent::__destruct()` method for that was deemed inappropriate. While convenient, there are restrictions on what can safely be relied on in destructor behavior.

arkaitzgarro: > I prefer not to rely on the middleware `terminate` because the documentation implies it is only call when using FastCGI:

How would you solve that using PHP-FPM then? The middleware stops the transaction and collects the events on `terminate`. Any event we could listen instead?

dstepe: The use of `Application::terminating()` appears to work in all cases I have tried. My testing is with Docker containers using PHP-FPM. I am not sure how a queue worker process might handle this, however. I noticed you already have some explicit sends in there since you could not rely on the `Agent::__destruct()`. I actually meant to bring this up and forgot. I think we need to so additional validation before you make a new release.

I'm also not convinced that I am interpreting the Laravel documentation on "Terminable Middleware" correctly. I quoted what is on the docs (https://laravel.com/docs/7.x/middleware#terminable-middleware) but have not tested exhaustively to understand the true behavior.

dstepe: I reread the article I linked to above. It suggests my interpretation of the docs is incorrect. The middleware `terminate` method is always called, even when not using FastCGI. The difference is that when using FastCGI (or PHP-FPM), the process explicitly flushes the response on the connection before proceeding to terminate.

The value of using the `Application::terminate()` is that it is called after all middleware termination is complete. That provides an opportunity for any other middleware to close transactions, etc before the `Agent::send()` is executed.

Let me know how this reads to you and I've missed anything.

I'm still not sure how this applies to long running jobs, but you seem to have done some work on that and I look forward to working on that as we have a number of queue workers and scheduled jobs that I would love to observe.

arkaitzgarro: To my understanding, the terminate method in the middleware should work regardless of the process manager indeed. It is called at the very end of the [index.php](https://github.com/laravel/laravel/blob/master/public/index.php#L55) file, so the actual structure should work as expected.

Same for jobs, as you pointed out, we have to manually call `$agent->send` because the class was never destroyed. It is much more cleaner now that we don't depend on magic methods.

Issue #67, opened 2020-09-06
Initializing the FrameworkCollector in the Agent creates a register/boot order issue


dstepe:
The `ServiceProvider` resolves the `Agent` class and calls `Agent::registerInitCollectors()` during the Laravel register phase.

https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/ServiceProvider.php#L93

Because this happens during the register phase of the Laravel boot process, other service providers may not have completed registering services that may be required by the `Agent` creation process.

The `FrameworkCollector` needs to load before the boot phase, however.

I suggest the framework collector object be created outside of the `Agent` class so it can register the necessary listeners, but that it then be given to the agent object after the agent is created. This should remove the need to create the `Agent` in the service container during the register phase.

I should have PR for this shortly.

PR #70, opened 2020-09-12
Chore/separate collector init from agent creation

dstepe:
There are two changes in this PR. First, the AgentBuilder class is now extended and given features specific to creating the Agent for this package. This is a much cleaner approach now supported by the underlying agent package.

Second, and more significantly, the process of registering and booting the agent object has been reworked to address #67 . The Agent class was previously resolved from the container during the register phase, which occurred before all service providers had run. I need to be able to influence the agent creation in my applications, specifically to disable tls cert verification, which used to be a configuration value. This is done in the project's AppServiceProvider, as shown in this example:

```php
    public function register()
    {
        $this->app->bind(AgentBuilder::class, function () {
            $builder = new AgentBuilder();

            $builder->withHttpClient(new \Http\Adapter\Guzzle6\Client(new Client(['verify' => false])));

            $builder->withPreCommitCallback(function (RequestInterface $request) {
                Log::info(sprintf('Pre commit url is: %s', $request->getUri()));
            });

            $builder->withPostCommitCallback(function (ResponseInterface $response) {
                Log::info(sprintf('Post commit response status: %s', $response->getStatusCode()));
                Log::debug($response->getBody()->getContents());
            });

            return $builder;
        });
    }
```

Creating the collectors outside of the agent and adding them later in the boot process solves this issue.

The above example also highlights other features enabled by the AgentBuilder that the application can take advantage of. You could chose to just bake those into this package, though, rather than having each application do this individually.

I also want to point out that the new Agent package does not include an HTTP client and a suitable implementation must be provided by the application. See: 

https://github.com/nipwaayoni/elastic-apm-php-agent/blob/master/docs/install.md#http-clientinterface

This may very well mean a breaking change for your users, but it should be solvable by requiring the additional packages. While inconvenient, this does broaden the compatibility of the Agent package.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a749594ca9021d523316c5b4d74e8a63",
+    "content-hash": "7f1dd6ac473eaede48231df321d1fcc7",
     "packages": [
         {
             "name": "doctrine/inflector",
@@ -736,16 +736,16 @@
         },
         {
             "name": "nipwaayoni/elastic-apm-php-agent",
-            "version": "v7.3.0",
+            "version": "v7.4.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/nipwaayoni/elastic-apm-php-agent.git",
-                "reference": "4707a43e7f80f846990721a5a704a3da85425901"
+                "reference": "2361c60d1149f591e0599335a2cb9eeb1f992070"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/nipwaayoni/elastic-apm-php-agent/zipball/4707a43e7f80f846990721a5a704a3da85425901",
-                "reference": "4707a43e7f80f846990721a5a704a3da85425901",
+                "url": "https://api.github.com/repos/nipwaayoni/elastic-apm-php-agent/zipball/2361c60d1149f591e0599335a2cb9eeb1f992070",
+                "reference": "2361c60d1149f591e0599335a2cb9eeb1f992070",
                 "shasum": ""
             },
             "require": {
@@ -785,7 +785,7 @@
                 }
             ],
             "description": "A php agent for Elastic APM v2 Intake API",
-            "time": "2020-09-06T17:38:05+00:00"
+            "time": "2020-09-12T10:58:04+00:00"
         },
         {
             "name": "paragonie/random_compat",
--- a/composer.lock
+++ b/composer.lock
@@ -4,37 +4,42 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a749594ca9021d523316c5b4d74e8a63",
+    "content-hash": "7f1dd6ac473eaede48231df321d1fcc7",
     "packages": [
         {
             "name": "doctrine/inflector",
-            "version": "1.3.1",
+            "version": "1.4.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/inflector.git",
-                "reference": "ec3a55242203ffa6a4b27c58176da97ff0a7aec1"
+                "reference": "4650c8b30c753a76bf44fb2ed00117d6f367490c"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/inflector/zipball/ec3a55242203ffa6a4b27c58176da97ff0a7aec1",
-                "reference": "ec3a55242203ffa6a4b27c58176da97ff0a7aec1",
+                "url": "https://api.github.com/repos/doctrine/inflector/zipball/4650c8b30c753a76bf44fb2ed00117d6f367490c",
+                "reference": "4650c8b30c753a76bf44fb2ed00117d6f367490c",
                 "shasum": ""
             },
             "require": {
-                "php": "^7.1"
+                "php": "^7.2 || ^8.0"
             },
             "require-dev": {
-                "phpunit/phpunit": "^6.2"
+                "doctrine/coding-standard": "^7.0",
+                "phpstan/phpstan": "^0.11",
+                "phpstan/phpstan-phpunit": "^0.11",
+                "phpstan/phpstan-strict-rules": "^0.11",
+                "phpunit/phpunit": "^7.0 || ^8.0 || ^9.0"
             },
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "1.3.x-dev"
+                    "dev-master": "2.0.x-dev"
                 }
             },
             "autoload": {
                 "psr-4": {
-                    "Doctrine\\Common\\Inflector\\": "lib/Doctrine/Common/Inflector"
+                    "Doctrine\\Common\\Inflector\\": "lib/Doctrine/Common/Inflector",
+                    "Doctrine\\Inflector\\": "lib/Doctrine/Inflector"
                 }
             },
             "notification-url": "https://packagist.org/downloads/",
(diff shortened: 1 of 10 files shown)


composer.lock
```diff
@@ -4,37 +4,42 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "74b66ab59b15e1b0093cbaa7d84ccfde",
+    "content-hash": "7f1dd6ac473eaede48231df321d1fcc7",
     "packages": [
         {
             "name": "doctrine/inflector",
-            "version": "1.3.1",
+            "version": "1.4.3",
```
arkaitzgarro: Composer updating all dependencies, a classic :) Could you only update `nipwaayoni/elastic-apm-php-agent`?

src/AgentBuilder.php
```diff
@@ -0,0 +1,62 @@
...
+
+class AgentBuilder extends NipwaayoniAgentBuilder
+{
+    /** @var RequestStartTime */
+    private $startTime;
+
+    /** @var Collection */
+    private $collectors;
+
+    public function withRequestStartTime(RequestStartTime $startTime): self
```
arkaitzgarro, line 24: Any specific reason to not use a more _traditional_ setter like `setRequestStartTime`? Not sure if `with*` is a Laravel convention.

src/Contracts/DataCollector.php
```diff
@@ -2,10 +2,19 @@
...
 
+use AG\ElasticApmLaravel\Agent;
 use Illuminate\Support\Collection;
 
 interface DataCollector
 {
+    /*
+     * This allows making a concrete Agent available to the collector after everything
+     * is booted. The JobCollector class makes extensive use of the Agent, unlike other
+     * collectors. I think long term, creating a separate "middleware" for jobs will
```
arkaitzgarro, line 13: We already provide a [Job Middleware](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Jobs/Middleware/RecordTransaction.php#L7), if this is what you mean. The job collector was developed for Laravel 5.5 compatibility.

src/ServiceProvider.php
```diff
@@ -104,21 +112,48 @@ protected function registerMiddleware(): void
...
+            $this->app->tag(FrameworkCollector::class, self::COLLECTOR_TAG);
+        }
+
+        if (false !== config('elastic-apm-laravel.spans.querylog.enabled')) {
+            // DB Queries collector
+            $this->app->tag(DBQueryCollector::class, self::COLLECTOR_TAG);
+        }
+
+        // Http request collector
+        if ('cli' !== php_sapi_name()) {
```
arkaitzgarro:
```suggestion
        if ($this->includeFrameworkEvents()) {
```

arkaitzgarro (commented):
> The above example also highlights other features enabled by the AgentBuilder that the application can take advantage of. You could chose to just bake those into this package, though, rather than having each application do this individually.
> 
> I also want to point out that the new Agent package does not include an HTTP client and a suitable implementation must be provided by the application. See:
> 
> https://github.com/nipwaayoni/elastic-apm-php-agent/blob/master/docs/install.md#http-clientinterface
> 
> This may very well mean a breaking change for your users, but it should be solvable by requiring the additional packages. While inconvenient, this does broaden the compatibility of the Agent package.

I will prefer very much to provide a package that users can start using right away, without caring about HTTP clients. Do you think it will be possible to include Guzzle, but at the same time provide the functionality to override the client at the same time?

```php
public function register()
{
    $builder = $this->app->make(AgentBuilder::class);
    $builder->withHttpClient(new \Http\Adapter\Guzzle6\Client(new Client(['verify' => false])));
}
```

composer.lock
dstepe: Certainly. Sorry about that. It may be a day before I have all of this done, though.

src/AgentBuilder.php
dstepe: I prefer the `with` nomenclature because of the builder pattern behavior. We're providing details to the builder for it to use when creating an `Agent`, not really setting something on the builder itself. The `with` methods also return an instance of the builder object to enable fluent chaining, which is not a normal setter behavior.

The Agent package is also not coupled to Laravel, so I do not follow Laravel specific conventions. It's used in a number of non-Laravel PHP projects, so the conventions should be framework agnostic. That will hidden somewhat by your service provider and facade.

src/Contracts/DataCollector.php
dstepe: I did see the job middleware. I was reacting more the difference between how HTTP transactions are collected and how jobs are collected. With the HTTP transaction, the collector only registers listeners and the middleware does all of the work. The job collector, on the other hand does all the interaction with the agent and other work. That couples the JobCollector to the Agent in ways that other Collectors are not.

My comment was not clear on that, mostly because my thinking on subject evolved after making it. I've had a chance to learn more of your package now.

src/ServiceProvider.php
dstepe: Missed that. Will change.

composer.lock
dstepe: Turns out I had more time now than expected. This should be addressed in the latest commits.

src/ServiceProvider.php
dstepe: Addressed in new commits. I opted for a slightly different solution which I believe expresses the conditions more clearly.

dstepe:
I wanted to also point out the use of tags in the service container to find the registered Collectors when adding them to the Agent. This is my first opportunity to use tags and I think it works really well. I did not add this ability to the documentation, but this would be an alternate way for anyone implementing their own collector to register it. 

Going further, we could even differentiate different groups of collectors with different tags and encapsulate all the logic in the package. For example, a user could tag a collector as being in the 'http group' and it would only be added by the service provider when processing an http request. The user does not have to make the decision, they just always register the collector.

arkaitzgarro:
> I wanted to also point out the use of tags in the service container to find the registered Collectors when adding them to the Agent. This is my first opportunity to use tags and I think it works really well. I did not add this ability to the documentation, but this would be an alternate way for anyone implementing their own collector to register it.
> 
> Going further, we could even differentiate different groups of collectors with different tags and encapsulate all the logic in the package. For example, a user could tag a collector as being in the 'http group' and it would only be added by the service provider when processing an http request. The user does not have to make the decision, they just always register the collector.

That could be a nice feature to have indeed, I didn't know about tagged services in Laravel. We could take a look to once we are done with the migration.

Any thoughts about my comment providing a default http client, but allowing users to override with their own choice?

dstepe:
Sorry. I forgot to respond regarding the http client. I think it does make sense for your package to "just work". We need to take care forcing dependencies in the underlying Agent package, but you can make different decisions when using it.

I wasn't completely clear earlier. You are not required to provide an actual http client object. You only need to have a suitable dependency available. The `Connector` class in the agent package will discover a suitable client:

https://github.com/nipwaayoni/elastic-apm-php-agent/blob/master/src/Middleware/Connector.php#L89

The reason to provide an actual object is when you need to customize it. The previous package explicitly created a new client object, and passed in configuration to it. While this worked to change settings such as the tls verification, you were pinned to version of the client dependency and unable to provide alternate implementations for testing.

As a package provider, you have two ways to approach this. First, you could defer the choice to the user of your package. This is what we did with the Laravel package I hope to deprecate in favor of this one. The install doc points out the need to provide a suitable implementation:

https://github.com/nipwaayoni/elastic-apm-laravel#install

The minimum required on the part of your user is:

```bash
composer require http-interop/http-factory-guzzle php-http/guzzle6-adapter
```
The docs then describe how to customize the client the connector will use, but only if you need:

https://github.com/nipwaayoni/elastic-apm-laravel/blob/master/docs/configuration.md#http-client-configuration

Your second option is to make the choice on behalf of your users. You would simply skip the installation decision and include those two dependencies in your own `composer.json`. The benefit is your users do not need to explicitly require them. The downside is that they must ensure their entire project is compatible with what you choose.

I believe that packages and libraries should make the fewest dependency choices possible for users. That ensures the widest possible compatibility. That does need to be balanced against the work users must do to use your package, which is something you as the author are better positioned to understand.

arkaitzgarro:
@dstepe I'm merging into my branch and will perform some tests in a running application.

dstepe:
@arkaitzgarro Sounds good. I've tested with a simple application I run locally for validating the agent and it seems good at a basic level. More testing is welcome.

I'm working on another release of the Agent package which should support transaction sample rate among other things. I'll be testing that with the this package, but won't start a PR until you are comfortable this work is ready to go to release.

Issue #69, opened 2020-09-08
On-the-fly enable


cykirsch:
We're interested in having APM only send events 10% of the time, and handling this via feature flags. https://github.com/checkr/flagr if you're curious.

I'm curious what you think about how to approach this. We could make our own custom service provider that extends yours, in order to make a flagr request prior to reading the provided configuration option. Do you see a clean way we could provide the ability to hook into the library instead of making a custom service provider?

cykirsch:
We use `php artisan cache` at deploy time so we can't do it inside the config file.

dstepe:
To clarify, are you interested in sending a specific 10% or simply a random 10% of the transaction events? I'm not sure how a feature flag would accomplish this.

I believe this is actually a needed enhancement of of the underlying APM Agent. I've suggested to the author of this package that we switch to an alternate release, and transaction sampling rate is already a requested feature enhancement:

https://github.com/nipwaayoni/elastic-apm-php-agent/issues/22

Is that what you are looking for?

cykirsch:
That's great information, thank you for sharing. Ideally that feature is exactly what we'd want, but without it actually being implemented we are stuck with the same as what you mentioned--just override `active` event though it is not optimal.

dstepe:
I believe you could extend `AG\ElasticApmLaravel\Agent` class and override the `send()` method. Then add condition which only calls `parent::send()` based on your given transaction sampling rate. I have not tried this, but I don't see why it would not work.

There are caveats, however. The Agent class does not currently expose any functionality to clear its collected events. That's done as part of calling `send()` and would therefore not happen when non-sampled transactions. If you have long running PHP processes which handle multiple transactions, this would result in all events being sent, but in batches.

Second, exception events would also only be sent based on your sampling rate, which may lead to undercounting or even missing exceptions. That's not n% of exceptions, it's n% of all events, which would make it easy to miss exceptions.

These are types of issues I would expect a proper feature in the underlying package to address, but if your primary concern is simply reduce the overall sampling, this simple solution may suffice for now.

dstepe:
I should also point out that the Python agent referred to in that issue implements sampling a little differently than what I suggested would achieve:

> By default, the agent will sample every transaction (e.g. request to your service). To reduce overhead and storage requirements, you can set the sample rate to a value between 0.0 and 1.0. We still record overall time and the result for unsampled transactions, but no context information, labels, or spans.

[...]

PR #71, opened 2020-09-14
Implement ApmConfigService class

cykirsch:
This provides a couple core benefits:
 - puts the isAgentDisabled logic in a single location instead of repeating it
 - makes the package more extensible in that the configuration gathering can be modified apart from the service provider

This is related to https://github.com/arkaitzgarro/elastic-apm-laravel/issues/69. It's not the ideal solution but to me it seems like they are still useful improvements.

I haven't quite finished all the tests; due to our difference in timezones I wanted to get this PR up tonight in hopes of some feedback before wrapping it up tomorrow (if you are interested in it moving forward).


arkaitzgarro:
Looking good, thanks! Do you think we should create a Facade to access the package configuration from the apps?

cykirsch:
@arkaitzgarro Tough to say, most of it is available via the framework facade and helper already. Adding it now would just allow access to the "resolved" values. That is, 

- the config as formatted for the agent--I don't see the need
- the isAgentDisabled()--Potentially useful in custom collectors, but if you're making a custom collector you might as well just inject the service 🤷‍♂️ 

It'll be a quick add if you want me to, but I'd personally do without until someone asks for it.

cykirsch:
Unfortunately I'm running into more issues in actual usage. In trying to extend, I'm still hitting the parent config because it gets called when we init the FrameworkCollector. This is an issue that would be solved by the AgentBuilder concept in https://github.com/arkaitzgarro/elastic-apm-laravel/pull/70. I'm still open to improving this and merging later but for now I think I'd consider it broken without #70.

arkaitzgarro:
Hey @cykirsch, any update on this change?

cykirsch:
No update. I can try to sneak in some time this week to see if it works better on the latest RC.

cykirsch:
I don't have plans for this right now. Feel free to tag me with any questions if implementing this in some way in the future.

PR #72, opened 2020-09-18
chore: migrate current Elastic PHP agent

### LINKED ISSUE #52: Package collaboration
Greetings! I see that you use the `philkra/elastic-apm-php-agent` package to provide the underlying APM service. That package has been seen much activity, and by mutual agreement with the author, I have created a replacement package to continue development (https://github.com/nipwaayoni/elastic-apm-php-agent). Philkra also published a Laravel APM package based on his Agent which I picked up as well (https://github.com/nipwaayoni/elastic-apm-laravel).

Would you consider switching to the new Agent package? There are some breaking changes to deal with, but I'm trying to keep improving it while watching the official Elastic APM Agent for PHP (https://github.com/elastic/apm-agent-php). I'd be happy to get feedback to benefit your Laravel package as well as the Agent.

I'd also like to consider how we might consolidate the Laravel APM packages. I like what I see in your implementation and I believe the community would benefit from collaboration.

I open to discussion.

### LINKED ISSUE #67: Initializing the FrameworkCollector in the Agent creates a register/boot order issue

arkaitzgarro:
The current Elastic PHP agent is not under maintenance anymore, and the author [agreed on transfering the ownership](https://github.com/philkra/elastic-apm-php-agent/issues/136#issuecomment-621883404) to an [organization](https://github.com/nipwaayoni/elastic-apm-php-agent). This PR contains the necessary changes to make this migration possible. Thanks to @dstepe who perfomed most of the work to make it happen.

It includes the following changes:
- Replace `philkra/elastic-apm-php-agent` with `nipwaayoni/elastic-apm-php-agent`.
- Refactor ServiceProvider:  the new Agent includes an AgentBuilder to facilitate constructing the agent. Use it to fulfill the Agent implementation in the service container.
- The Agent no longer relies on the class `__destruct()` method to call `send()`. We need to explicitly call `send()` on the agent which is best done in an `App::terminating()` registered callback. This is executed after the framework has sent the response to the consumer.
- Remove injection of Agent class into `EventDataCollector` classes.
- The `FrameworkCollector` is created during the Laravel 'register' phase, which was forcing the Agent to be constructed as well. This occurred before other service providers had completed registering and therefore prevented modifying the Agent construction through the AgentBuilder abstraction.
- Collectors are now tagged during 'register' and given to the Agent, rather than being created within the Agent.
- The Agent is given to the Collector objects during the add process to meet the need of the `JobCollector` interacting with it.

Because of the changes to the Agent, I think we should release a version 2 of the package. If any customer is extending the existing implementation, this new version will break their code.

Closes #52, closes #67

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,9 @@
         "illuminate/routing": "^5.5|^6|^7",
         "illuminate/support": "^5.5|^6|^7",
         "jasny/dbquery-mysql": "^2.0",
-        "nipwaayoni/elastic-apm-php-agent": "^7.4"
+        "nipwaayoni/elastic-apm-php-agent": "^7.4",
+        "http-interop/http-factory-guzzle": "^1.0",
+        "php-http/guzzle6-adapter": "^2.0"
     },
     "require-dev": {
         "codeception/codeception": "^4.1",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "illuminate/routing": "^5.5|^6|^7",
         "illuminate/support": "^5.5|^6|^7",
         "jasny/dbquery-mysql": "^2.0",
-        "nipwaayoni/elastic-apm-php-agent": "^7.2"
+        "nipwaayoni/elastic-apm-php-agent": "^7.3"
     },
     "require-dev": {
         "codeception/codeception": "^4.1",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "illuminate/routing": "^5.5|^6|^7",
         "illuminate/support": "^5.5|^6|^7",
         "jasny/dbquery-mysql": "^2.0",
-        "nipwaayoni/elastic-apm-php-agent": "^7.3"
+        "nipwaayoni/elastic-apm-php-agent": "^7.4"
     },
     "require-dev": {
         "codeception/codeception": "^4.1",
--- a/composer.json
+++ b/composer.json
@@ -30,12 +30,6 @@
         "dms/phpunit-arraysubset-asserts": "^0.1.0",
         "codeception/mockery-module": "^0.4.0"
     },
-    "repositories": [
-        {
-            "type": "vcs",
-            "url": "https://github.com/Nipwaayoni/elastic-apm-php-agent"
-        }
-    ],
     "autoload": {
         "psr-4": {
             "AG\\ElasticApmLaravel\\": "src/"
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,8 @@
         "illuminate/http": "^5.5|^6|^7",
         "illuminate/routing": "^5.5|^6|^7",
         "illuminate/support": "^5.5|^6|^7",
-        "philkra/elastic-apm-php-agent": "dev-master",
-        "jasny/dbquery-mysql": "^2.0"
+        "jasny/dbquery-mysql": "^2.0",
+        "nipwaayoni/elastic-apm-php-agent": "^7.2"
     },
     "require-dev": {
         "codeception/codeception": "^4.1",
@@ -33,7 +33,7 @@
     "repositories": [
         {
             "type": "vcs",
-            "url": "https://github.com/philkra/elastic-apm-php-agent"
+            "url": "https://github.com/Nipwaayoni/elastic-apm-php-agent"
         }
     ],
     "autoload": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "7f1dd6ac473eaede48231df321d1fcc7",
+    "content-hash": "fbfc852c8e05edf0947f144eeb388bc7",
     "packages": [
         {
             "name": "doctrine/inflector",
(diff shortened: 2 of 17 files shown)


arkaitzgarro:
@dstepe did you have time to test the changes on your side? Everything works as expected except for asynchronous jobs. Looks like the collector is registered, but callbacks on events are never executed. Not sure if it's because of service tags or the way workers boot/register services. Or both. I couldn't find the reason.

dstepe:
@arkaitzgarro I'm very happy to see this moving forward. I've been testing on a small demo app I maintain for this purpose. I'll add jobs to that and have a go at it. I'll start adding this to some of my production applications next week.

dstepe:
@arkaitzgarro I suggest holding off on a new release of this package until I get the next release of the Agent package out. I'm adding things which I believe users of this package will find useful and it may make sense to bundle them all as one major release. I expect to finish that in the next few days.

dstepe:
@arkaitzgarro The issue with jobs not being recorded is due to the work I did to move the `Agent` creation out of the Laravel register phase. The previous register process resulted in the agent being created and adding all the collectors, including the job collector, during the register phase. Now that I split that up, the agent is only resolved when something needs it. In the case of http requests, the http middleware asks for an `Agent` to be provided in the constructor. That happens early and results in everything working.

There is currently nothing which causes an `Agent` to made before a job is processed, therefore the `JobCollector` object is never created to register its listeners.

I can force everything to work by resolving the `Agent` at the end of the `ServiceProvider::boot()` method, but that seems ugly. Now that I know the problem, I'll track down a more elegant solution.

Is there a reason you don't automatically register the job middleware as you do the http middleware? If we did so, and the job middleware asked for an `Agent`, I think we would get the same results as with the http requests.

dstepe:
Ah. After reading a bit more I see that job middleware is different. It need to be specified in the job, as stated in your README. I'm sure we can find a way to make this work without complicating the application job setup.

dstepe:
@arkaitzgarro I submitted a PR to the working branch that should enable the job event collection. I still feel there is an opportunity in the JobCollector class. It depends on an Agent, unlike all the other collectors, which makes the problem awkward to solve. But that can be handled as internal refactoring later. I think this solution works with no change for users of the package.

dstepe:
This will close #52  and #67 if you want to link them.

dstepe:
@arkaitzgarro What's the status of this PR? Is there anything I can do to assist moving it forward?

arkaitzgarro:
> @arkaitzgarro What's the status of this PR? Is there anything I can do to assist moving it forward?

Hey @dstepe, I was actually waiting for your feedback, since you mentioned that you were planing to use this implementation in your production apps. From my side, this is ready to go and can be merged into `master` and we could release a `v2.0-RC1`. I would also like to include [these extra collectors](https://github.com/arkaitzgarro/elastic-apm-laravel/pull/76) in the new version.

Since this will introduce breaking changes, I'm also thinking on dropping support for Laravel 5.5, but this can be discussed in a separate thread.

dstepe:
Sorry about that. I have on particular app I want to get this into and I ended up pulled onto something else last week. I'll get back to it in the next day or two. The package has worked well in the cases I have tried, however, and I'm confident in it at this point.

What changes are leading to dropping support for Laravel 5.5? Not that I object, but I'm curious.

arkaitzgarro:
> Sorry about that. I have on particular app I want to get this into and I ended up pulled onto something else last week. I'll get back to it in the next day or two. The package has worked well in the cases I have tried, however, and I'm confident in it at this point.

Great, let me know when we are ready to go!

> What changes are leading to dropping support for Laravel 5.5? Not that I object, but I'm curious.

The main reason is the introduction of a scheduled command collector, as `Illuminate\Console\Events\ScheduledTask*` events are only available from Laravel 5.8 version.

dstepe:
Would it make sense to have the scheduled command collector only register listeners when the version of Laravel is at or above 5.8 then? I know we have some small internal apps that have not been updated to 5.8 or higher yet, and the absence of scheduled command collecting would not impact them. I'd still like to keep up with other feature releases of this package though.

arkaitzgarro:
Definitely. I’ve just found out the existence of `version_compare` PHP function, so we could use that to dynamically register the collector or the listeners and keep the compatibility with 5.5.

dstepe:
@arkaitzgarro How would you like to handle incorporating the latest code from the main branch into this PR? I could rebase my branch, but I'm not sure how that will work with the PR. We could simply make the changes directly, or cherry pick the commits, but then you'll likely end up with conflicts on the final merge. Do you have a preference?

arkaitzgarro:
Hey @dstepe, I rebase master into the branch, it is up to date now.

dstepe:
@arkaitzgarro I've had this branch running in a production app for a couple of days now and not seen any problem. I think it's good to go.

For release planning, I suggest this will be a good RC1, but not complete. I'm preparing a 7.5 release of the Agent package and while I don't expect any breaking changes, there are new features and deprecations that would make sense to deal with as part of a new major release in this package. I should have that release done and be able to start PRs to you in a few days. I've been holding off until this is merged since it's gotten quite large on it's own.

arkaitzgarro:
> @arkaitzgarro I've had this branch running in a production app for a couple of days now and not seen any problem. I think it's good to go.
> 
> For release planning, I suggest this will be a good RC1, but not complete. I'm preparing a 7.5 release of the Agent package and while I don't expect any breaking changes, there are new features and deprecations that would make sense to deal with as part of a new major release in this package. I should have that release done and be able to start PRs to you in a few days. I've been holding off until this is merged since it's gotten quite large on it's own.

That's really good news! I'm merging this PR and creating a `v2.0-RC1` release.

PR #73, opened 2020-09-19
Make Agent during boot when running in cli

dstepe:
This should re-enable the collection of events during async job processing. However, I've not used the job metrics before and I'm not sure what I'm seeing in APM is correct.


arkaitzgarro:
Works as expected, thanks!

PR #74, opened 2020-09-20
fix: transaction and span times for async jobs

arkaitzgarro:
This fixed an issue with timing on asynchronous jobs. We depend on `RequestStartTime` to calculate when spans start in relation to the transaction, but for async jobs, this class holds the time when the queue was started, not when the transaction was started, showing an incorrect timeline for jobs.

I created a new method that allows overriding the `$start_time` value, which is called when a job starts processing, but only for async queues. That way, every time a job is processed, we can set the correct starting time.

Bonus points:
- Use Laravel's `runningInConsole` method to detect if we are running the code in the console, instead of depending on `php_sapi_name` output.
- Use snake_case notation for variable names. We follow this pattern, even though is not mentioned anywhere.

@dstepe could you take a look?

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -2,7 +2,6 @@
 
 namespace AG\ElasticApmLaravel;
 
-use AG\ElasticApmLaravel\Collectors\RequestStartTime;
 use AG\ElasticApmLaravel\Contracts\DataCollector;
 use AG\ElasticApmLaravel\Events\LazySpan;
 use Illuminate\Support\Collection;
@@ -28,19 +27,16 @@ use Nipwaayoni\Stores\TransactionsStore;
 class Agent extends NipwaayoniAgent
 {
     protected $collectors;
-    protected $request_start_time;
 
     public function __construct(
         Config $config,
-        ContextCollection $sharedContext,
+        ContextCollection $shared_context,
         Connector $connector,
-        EventFactoryInterface $eventFactory,
-        TransactionsStore $transactionsStore,
-        RequestStartTime $startTime
+        EventFactoryInterface $event_factory,
+        TransactionsStore $transactions_store
     ) {
-        parent::__construct($config, $sharedContext, $connector, $eventFactory, $transactionsStore);
+        parent::__construct($config, $shared_context, $connector, $event_factory, $transactions_store);
 
-        $this->request_start_time = $startTime->microseconds();
         $this->collectors = new Collection();
     }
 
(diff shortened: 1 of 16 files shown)


src/Agent.php
```diff
@@ -28,19 +27,16 @@
...
-    protected $request_start_time;
 
     public function __construct(
         Config $config,
-        ContextCollection $sharedContext,
+        ContextCollection $shared_context,
         Connector $connector,
-        EventFactoryInterface $eventFactory,
-        TransactionsStore $transactionsStore,
-        RequestStartTime $startTime
```
dstepe, line 39: Just to confirm, the start time is being removed from the `Agent` class because it will now be provided directly to the `EventDataCollector` classes?

src/AgentBuilder.php
```diff
@@ -37,20 +37,20 @@ public function withEventCollectors(Collection $collectors): self
...
+        if (null === $this->start_time) {
+            $this->start_time = new RequestStartTime(microtime(true));
         }
 
         if (null === $this->collectors) {
             $this->collectors = new Collection();
         }
 
-        $agent = new Agent($config, $sharedContext, $connector, $eventFactory, $transactionsStore, $this->startTime);
+        $agent = new Agent($config, $shared_context, $connector, $event_factory, $transactions_store, $this->start_time);
```
dstepe: The start time was removed from the `Agent` class so it should no longer be passed in the constructor here.

That also means all of the start time related code in the `AgentBuilder` can be removed.

dstepe (commented):
Looks pretty good to me other than the couple of comments. I'll do more testing with the `migrate-php-agent` branch once this is merged in.

Sorry I overlooked the naming convention. I should have caught that and followed your standard.

src/Agent.php
arkaitzgarro: Correct, everywhere the start time is needed, it will be injected from now on.

src/AgentBuilder.php
arkaitzgarro: Good catch, I will remove the references 👍

dstepe: You'll need to remove that method call from the `ServiceProvider` as well:

https://github.com/arkaitzgarro/elastic-apm-laravel/blob/fix/async-job-start-time/src/ServiceProvider.php#L99

dstepe: And this is no longer needed in that case:

https://github.com/arkaitzgarro/elastic-apm-laravel/blob/fix/async-job-start-time/src/ServiceProvider.php#L91

arkaitzgarro: Remove and commit the changes, because I somehow forgot about it 🤦

PR #76, opened 2020-09-21
feat(collectors): add command and scheduled tasks collectors

### LINKED ISSUE #75: Add a collector for artisan commands
Add a new collector that listens to `CommandStarting` and `CommandFinished` events, so we can measure Laravel's artisan commands.

https://laravel.com/api/5.8/Illuminate/Console/Events/CommandStarting.html
https://laravel.com/api/5.8/Illuminate/Console/Events/CommandFinished.html

`ArtisanStarting` event could be interesting to be included as well https://laravel.com/api/5.8/Illuminate/Console/Events/ArtisanStarting.html

countless-integers:
Adds collectors for cli commands and scheduled tasks.

Closes #75 

todo: 

* [x] tests <sup>[1](#footnote)</sup>
* [x] extra metadata (especially for scheduled tasks) (?)

<a name="footnote">1</a>: not sure what to do with tests for the scheduler the classes are missing in L5.5 + some issues with starting test  in command collector

### DIFF
--- a/src/Collectors/CommandCollector.php
+++ b/src/Collectors/CommandCollector.php
@@ -8,7 +8,6 @@ use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Console\Events\CommandStarting;
 use Illuminate\Support\Facades\Log;
 use PhilKra\Events\Transaction;
-use PhilKra\Exception\Transaction\UnknownTransactionException;
 use Throwable;
 
 /**
@@ -50,15 +49,6 @@ class CommandCollector extends EventDataCollector implements DataCollector
                 }
             }
         });
-    }
-
-    protected function getTransaction(string $transaction_name): ?Transaction
-    {
-        try {
-            return $this->agent->getTransaction($transaction_name);
-        } catch (UnknownTransactionException $e) {
-            return null;
-        }
     }
 
     protected function startTransaction(string $transaction_name): Transaction
@@ -115,12 +105,5 @@ class CommandCollector extends EventDataCollector implements DataCollector
         }
 
         return $this->shouldIgnoreTransaction($transaction_name) ? '' : $transaction_name;
-    }
-
-    protected function shouldIgnoreTransaction(string $transaction_name): bool
-    {
-        $pattern = $this->config->get('elastic-apm-laravel.transactions.ignorePatterns');
-
-        return $pattern && preg_match($pattern, $transaction_name);
     }
 }
--- a/src/Collectors/CommandCollector.php
+++ b/src/Collectors/CommandCollector.php
@@ -45,7 +45,10 @@ class CommandCollector extends EventDataCollector implements DataCollector
             if ($transaction_name) {
                 $transaction = $this->getTransaction($transaction_name);
                 if ($transaction) {
-                    $this->stopTransaction($transaction_name, 200);
+                    $this->stopTransaction(
+                        $transaction_name,
+                        $event->exitCode === 0 ? 200 : 500
+                    );
                     $this->send($event);
                 }
             }
--- a/src/Collectors/CommandCollector.php
+++ b/src/Collectors/CommandCollector.php
@@ -3,6 +3,7 @@
 namespace AG\ElasticApmLaravel\Collectors;
 
 use AG\ElasticApmLaravel\Contracts\DataCollector;
+use Carbon\Carbon;
 use GuzzleHttp\Exception\ClientException;
 use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Console\Events\CommandStarting;
@@ -35,8 +36,8 @@ class CommandCollector extends EventDataCollector implements DataCollector
                 return;
             }
 
-            $this->startTransaction($transaction_name);
-            $this->setTransactionType($transaction_name);
+            $transaction = $this->startTransaction($transaction_name);
+            $this->addMetadata($transaction);
         });
 
         $this->app->events->listen(CommandFinished::class, function (CommandFinished $event) {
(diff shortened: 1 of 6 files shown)


src/Collectors/ScheduledTaskCollector.php
```diff
@@ -0,0 +1,133 @@
...
+class ScheduledTaskCollector extends EventDataCollector implements DataCollector
+{
+    public function getName(): string
+    {
+        return 'scheduled-task-collector';
+    }
+
+    public function registerEventListeners(): void
+    {
+        $this->app->events->listen(ScheduledTaskStarting::class, function (ScheduledTaskStarting $event) {
```
arkaitzgarro, line 27: If I'm not wrong `ScheduledTask*` events were introduced in Laravel 5.8, while we support version 5.5. Can we register the listeners only if Laravel version is `>= 5.8`?

countless-integers: I've looked for it in Laravel repos and I only see those classes in 6+ 😬 . I could add a framework version check or just a class existence check (theoretically one could implement that class if they really wanted in older versions).

arkaitzgarro: Yes sorry, events for commands are the ones introduced in Laravel 5.8. We'll need checks for both collectors then.

src/Collectors/CommandCollector.php
```diff
@@ -0,0 +1,126 @@
...
+        $transaction_name = $event->command;
+
+        if (null === $transaction_name) {
+            return '';
+        }
+
+        return $this->shouldIgnoreTransaction($transaction_name) ? '' : $transaction_name;
+    }
+
+    protected function shouldIgnoreTransaction(string $transaction_name): bool
```
arkaitzgarro: Looks like we have several methods duplicated across collectors, could you maybe extract them to `EventDataCollector`?

src/Collectors/ScheduledTaskCollector.php
countless-integers: [`CommandStarting`](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Console/Events/CommandStarting.php) and [`CommandFinished`](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Console/Events/CommandFinished.php) are both in 5.5 though.

src/Agent.php
```diff
@@ -2,14 +2,17 @@
...
 use AG\ElasticApmLaravel\Collectors\DBQueryCollector;
 use AG\ElasticApmLaravel\Collectors\FrameworkCollector;
 use AG\ElasticApmLaravel\Collectors\HttpRequestCollector;
 use AG\ElasticApmLaravel\Collectors\JobCollector;
+use AG\ElasticApmLaravel\Collectors\ScheduledTaskCollector;
 use AG\ElasticApmLaravel\Collectors\SpanCollector;
 use AG\ElasticApmLaravel\Contracts\DataCollector;
 use AG\ElasticApmLaravel\Events\LazySpan;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\App;
```
arkaitzgarro: Is this facade being used somewhere?

countless-integers: nope, it should not be there.

countless-integers:
It gets interesting when a command fails though. I'm missing some way to report the error :/ . Perhaps some global exception listener in the framework collector (if it's not there already and I'm just no seeing it, I have a vague recollection that I've seen errors reported before...)

Screenshot 2020-10-01 at 14 51 46

arkaitzgarro:
> It gets interesting when a command fails though. I'm missing some way to report the error :/ . Perhaps some global exception listener in the framework collector (if it's not there already and I'm just no seeing it, I have a vague recollection that I've seen errors reported before...)

`CommandFinished` event includes a `$exitCode` property. Can it be that the event is fired for both successful and failed commands? https://github.com/laravel/framework/blob/43bea00fd27c76c01fd009e46725a54885f4d2a5/src/Illuminate/Console/Application.php#L93

countless-integers:
> > It gets interesting when a command fails though. I'm missing some way to report the error :/ . Perhaps some global exception listener in the framework collector (if it's not there already and I'm just no seeing it, I have a vague recollection that I've seen errors reported before...)
> 
> `CommandFinished` event includes a `$exitCode` property. Can it be that the event is fired for both successful and failed commands? https://github.com/laravel/framework/blob/43bea00fd27c76c01fd009e46725a54885f4d2a5/src/Illuminate/Console/Application.php#L93

I'd expect that. I'll see if I can reproduce it and adjust the "fake status code" from 200 to something else. If I remember correctly, in Symfony, the command could return a status code in the handle method and that was it. 

However what I meant was when an exception is thrown and not handled in the command, then listener doesn't seem to fire, but the app terminate middleware executes send on the agent. In phikra there was [this naughty thing](https://github.com/philkra/elastic-apm-laravel/blob/72855430638f4011f3e18d4daf61ff3526310363/src/Exceptions/ApmAgent.php) that adding errors.

countless-integers:
Ok, I've added the status code check. 

I've also noticed I've been wrong about `CommandFinished` -- it doesn't seem to fire when there's an unhandled exception thrown in the command. Terminate middleware still sends data to APM though.

src/Collectors/CommandCollector.php
```diff
@@ -47,7 +47,7 @@ public function registerEventListeners(): void
                 if ($transaction) {
                     $this->stopTransaction(
                         $transaction_name,
-                        $event->exitCode === 0 ? 200 : 500
+                        0 === $event->exitCode ? 200 : 500
```
arkaitzgarro: Are you not interested in the real exit code value? I think it might be useful for debugging purposes.

```suggestion
                        $event->exitCode
```

arkaitzgarro: I'm starting to believe that GH code suggestions are not the best idea, they only fix an isolated part of the code.

arkaitzgarro (approved):

PR #77, opened 2020-09-27
Reset EventDataCollectors and add Metadata after each send

### LINKED ISSUE #57: did not recognize object type
Hello,

When logging laravel jobs, I have this kind of error : 
```
production.ERROR: Client error: `POST /intake/v2/events` resulted in a `400 Bad Request` response:
 "errors": [
 {
 "message": "did not recognize object type", 
```

I'm using laravel 7.4
php 7.4
Kibana 7.6.2

Thank you !

dstepe:
Closes #57 

The current event collection results in problems with long running worker processes. The required metadata is not being added before each batch and the event collectors are not being cleared after sending. Overriding the base `Agent::send()` allows us to work around both issues as a short term fix. The base `Agent` package will improve the event collection process in a future release to address this. You may want to look into a better way to reset the collectors in this package.

I have tested this on my reference application with a worker queue and it appears to work as expected. However, I do not have the philkra agent package installed and cannot attest to this being fully ready to merge. I offer is a starting point for addressing the reported issue.

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -120,7 +120,7 @@ class Agent extends PhilKraAgent
          * worker processes. A future release of the Agent package should handle event
          * collection better and remove the need for this.
          */
-        $this->putEvent(new Metadata([], $this->getConfig(), $this->agentMetadata()));
+        $this->putEvent(new Metadata([], $this->getConfig()));
 
         return $sent;
     }
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -3,6 +3,7 @@
 namespace AG\ElasticApmLaravel;
 
 use AG\ElasticApmLaravel\Collectors\DBQueryCollector;
+use AG\ElasticApmLaravel\Collectors\EventDataCollector;
 use AG\ElasticApmLaravel\Collectors\FrameworkCollector;
 use AG\ElasticApmLaravel\Collectors\HttpRequestCollector;
 use AG\ElasticApmLaravel\Collectors\JobCollector;
@@ -11,6 +12,7 @@ use AG\ElasticApmLaravel\Contracts\DataCollector;
 use AG\ElasticApmLaravel\Events\LazySpan;
 use Illuminate\Support\Collection;
 use PhilKra\Agent as PhilKraAgent;
+use PhilKra\Events\Metadata;
 
 /**
  * The Elastic APM agent sends performance metrics and error logs to the APM Server.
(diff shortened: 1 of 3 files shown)


src/Agent.php
```diff
@@ -102,4 +104,17 @@ public function getRequestStartTime(): float
...
+
+    public function send(): bool
+    {
+        $sent = parent::send();
+
+        $this->collectors->each(function (EventDataCollector $collector) {
+            $collector->reset();
+        });
+
+        $this->putEvent(new Metadata([], $this->getConfig(), $this->agentMetadata()));
```
arkaitzgarro: You are adding this `Metadata` event after previous events have been sent to make it available for the next "connection" right?

src/Collectors/EventDataCollector.php
```diff
@@ -136,4 +136,10 @@ private function toMilliseconds(float $time): float
     {
         return round($time * 1000, 3);
     }
+
+    public function reset(): void
```
arkaitzgarro, line 140: Could you add this method to the `AG\ElasticApmLaravel\Contracts\DataCollector` interface?

src/Agent.php
dstepe: Yes. That reproduces the state of the Collector as it would immediately after the Agent is constructed. It's not what I consider an elegant solution, but it is completely within the context of this package and should work with the old philkra agent, as well as the 7.4 release of the nipwaayoni agent. I don't think I'll fix this in 7.5, but I intend to redo the event creation/collection in the next major release. I added a comment to explain why this code exists.

src/Collectors/EventDataCollector.php
dstepe: Done. Waiting on the checks to confirm. I rebased and squashed the commits to clean this up as well.

src/Agent.php
arkaitzgarro: @dstepe keep in mind that this fix is for the previous agent, and `agentMetadata` method is not available :)

src/Agent.php
```diff
@@ -102,4 +104,24 @@ public function getRequestStartTime(): float
...
+            $collector->reset();
+        });
+
+        /*
+         * Push new metadata onto the stack in preparation for the next send event. This
+         * simulates the behavior when a new Agent is created and is needed for long running
+         * worker processes. A future release of the Agent package should handle event
+         * collection better and remove the need for this.
+         */
+        $this->putEvent(new Metadata([], $this->getConfig(), $this->agentMetadata()));
```
arkaitzgarro:
```suggestion
        $this->putEvent(new Metadata([], $this->getConfig()));
```

dstepe: Fixed. I actually installed the old package this time. Should be good. Sorry about that.

cykirsch:
@dstepe This fix was perfect, I really appreciate you handling it so quickly. 🎉

PR #78, opened 2020-09-30
Chore/support elastic apm env config

dstepe:
I'm adding support for the 7.5 release of the Agent (https://github.com/nipwaayoni/elastic-apm-php-agent/releases/tag/v7.5.0). I don't expect breaking changes in the Laravel integration, but there will be what I hope are positive differences.

One noticeable change is how the Agent can now be configured. Almost all configuration options can now be set by ELASTIC_APM_* environment variables. I wanted to do this after implementing the Transaction Sample Rate feature and realizing I had to do work on this package to do nothing more than pass that setting into the agent. All the other APM agents support environment variables for configuration and it makes sense to so here as well. That way your users can directly configure the agent features without you need to do extra work.

The previous mechanism of passing options into the Config constructor is still supported, and takes precedence over env values, so there should be no required change for users of this package. However, the package itself needs to be changed to accommodate anyone wanting to move to env variables. 

All default values have been removed from the config file so the Agent package can correctly determine when intentional Config arguments are given. The defaults in the Agent package should be consistent with the previous defaults in this config, so your users should not see a change. The ServiceProvider is also updated to filter out empty values before passing to the Config constructor.

Related to using environment variables, we can no longer set a default application name in the config because that would always override the environment variable if the user tried to set it that way. There is a new "defaultServiceName" configuration option which will be used if the value is not otherwise provided.

You can now pass a PSR-3 compliant logger object into the configuration (this obviously can't be done through an environment variable). This allows better insight into the internals of the agent objects. Log messages are still evolving in those classes, but it's a good start. You can set the log level for that package independently of your Laravel app log level to reduce the noise if you are not trying to debug the agent.

I believe I have addressed the issues which led you to create the LazySpan class. I have removed that class from this package and switched to using the base Span. The implementation of Transaction Sample Rate caused some changes in the classes and it is much better if you can stick with the base classes if possible.

I added methods to track the current transaction in the Agent. I needed this to properly support distributed tracing (which I am really excited about). When making a new request I want to trace within the Laravel app, I need the trace ID from the current transaction to include in the header.

I have not started updating documentation in this project yet. I wanted to get the code and feature changes to you for discussion first. I'll start on docs next and leave this is a draft PR while you review the code.


dstepe:
As I'm working through the docs, I realize that there is one change in the default configuration. The `APM_BACKTRACEDEPTH` (now `ELASTIC_APM_STACK_TRACE_LIMIT`) was defaulted to 25, but the Agent package defaults to 0 (unlimited).

arkaitzgarro:
> I'm adding support for the 7.5 release of the Agent (https://github.com/nipwaayoni/elastic-apm-php-agent/releases/tag/v7.5.0). I don't expect breaking changes in the Laravel integration, but there will be what I hope are positive differences.

Great job once again, thanks for the support.

> One noticeable change is how the Agent can now be configured. Almost all configuration options can now be set by ELASTIC_APM_* environment variables. I wanted to do this after implementing the Transaction Sample Rate feature and realizing I had to do work on this package to do nothing more than pass that setting into the agent. All the other APM agents support environment variables for configuration and it makes sense to so here as well. That way your users can directly configure the agent features without you need to do extra work.

I saw that you added the configuration values to the documentation. I think it's good that users can configure the underlying package directly, they can immediately benefit from your features without us needing to implement a middle layer.

> The previous mechanism of passing options into the Config constructor is still supported, and takes precedence over env values, so there should be no required change for users of this package. However, the package itself needs to be changed to accommodate anyone wanting to move to env variables.
> 
> All default values have been removed from the config file so the Agent package can correctly determine when intentional Config arguments are given. The defaults in the Agent package should be consistent with the previous defaults in this config, so your users should not see a change. The ServiceProvider is also updated to filter out empty values before passing to the Config constructor.
>
> Related to using environment variables, we can no longer set a default application name in the config because that would always override the environment variable if the user tried to set it that way. There is a new "defaultServiceName" configuration option which will be used if the value is not otherwise provided.
> 
> You can now pass a PSR-3 compliant logger object into the configuration (this obviously can't be done through an environment variable). This allows better insight into the internals of the agent objects. Log messages are still evolving in those classes, but it's a good start. You can set the log level for that package independently of your Laravel app log level to reduce the noise if you are not trying to debug the agent.
> 
> I believe I have addressed the issues which led you to create the LazySpan class. I have removed that class from this package and switched to using the base Span. The implementation of Transaction Sample Rate caused some changes in the classes and it is much better if you can stick with the base classes if possible.

Good, I will give it a try. I see the timing is handled in a different way now, I will make sure spans are shown as expected.

> I added methods to track the current transaction in the Agent. I needed this to properly support distributed tracing (which I am really excited about). When making a new request I want to trace within the Laravel app, I need the trace ID from the current transaction to include in the header.

This is really exiting indeed!

> I have not started updating documentation in this project yet. I wanted to get the code and feature changes to you for discussion first. I'll start on docs next and leave this is a draft PR while you review the code.

arkaitzgarro:
> As I'm working through the docs, I realize that there is one change in the default configuration. The `APM_BACKTRACEDEPTH` (now `ELASTIC_APM_STACK_TRACE_LIMIT`) was defaulted to 25, but the Agent package defaults to 0 (unlimited).

I will add it to the breaking changes when we make the final release, thanks for letting me know.

README.md
```diff
@@ -136,6 +153,39 @@ public function boot()
...
+    $this->app->tag(MailMessageCollector::class, \AG\ElasticApmLaravel\ServiceProvider::COLLECTOR_TAG);
+}
+```
+
+## Distributed Tracing
+
+[Distributed tracing](https://www.elastic.co/guide/en/apm/get-started/current/distributed-tracing.html) allows you to associate transactions in your Laravel application with transactions in services your application consumes. For example, if your Laravel application calls REST resources as part of handling a request, the REST transaction details will appear within your application transaction in Elastic APM. You enable distributed tracing by including an appropriate header in the http request your application makes to another service. For example:
+
+```php
+$transaction = $agent->currentTransaction();
```
arkaitzgarro: Should we create an Agent facade? Otherwise user will need to get the instance like `app(Agent::class);` every time they need access to the Agent.

src/Middleware/RecordTransaction.php
```diff
@@ -50,6 +50,7 @@ public function handle(Request $request, Closure $next)
 
         // Start a new transaction
         $transaction = $this->startTransaction($transaction_name);
+        $this->agent->setCurrentTransaction($transaction);
```
arkaitzgarro: Are we interested in distributed transaction in jobs? If so, we will need add this to the JobCollector as well.

src/Agent.php
```diff
@@ -61,18 +65,28 @@ public function getCollector(string $name): DataCollector
         return $this->collectors->get($name);
     }
 
+    public function setCurrentTransaction(Transaction $transaction): void
```
arkaitzgarro, line 82: Do we need to unset the transaction once is sent? Thinking about async jobs here.

README.md
dstepe: Yes, that makes sense. I sort of have philosophical issues with facades, but I understand their value to the general Laravel community. I'll take a crack at adding it. That would also be a good place to put at least one over helper method I want to add.

src/Middleware/RecordTransaction.php
dstepe: We need to be aware of the impact on long running workers, but since this is in the http middleware, the current transaction actually would not be set for a job. That raises an entirely different problem to address since the job should also be able to access the current transaction for distributed tracing. I'll make a note of this and work it into the job middleware.

dstepe: Grr. I thought was replying to the next comment. That's what I get for readying comments in email first.

Yes, you are exactly correct and I'll work on addressing this.

README.md
dstepe: I've added an `ApmAgent` facade and exposed two methods for helping with the distributed tracing header.

src/Agent.php
dstepe: Yes. I added this to the `Agent::send()` method which we had previous overridden anyway. That seems to be the best place to hook this in.

dstepe:
@arkaitzgarro This should address the outstanding issues. I want to spend some time writing tests. I've not been good about using TDD and need to learn more about codeception.

I also updated the stack trace limit default to 50 to be consistent with what it looks like other agents are doing:

https://docs.google.com/spreadsheets/d/1JJjZotapacA3FkHc2sv_0wiChILi3uKnkwLTjtBmxwU/edit#gid=0

dstepe:
@arkaitzgarro I have couple of questions not directly related to this PR, but not quite worth opening a new issue.

1. What is your expectation for resolving conversations on a PR? I prefer that all conversations be resolved for merging, but I also expect that the person who opened the conversation should be the one to resolve it. That is an indication that the original concern has been addressed. (With the exception of a contributor who is not responding in a timely way and delaying a release.)

2. I noticed that the `config/ApmConfigTest.php` is not being run when following your documented development steps for unit tests. Is that intentional and if so why? Or should that test class be moved into the `unit` directory and made part of the standard unit suite?

3. Do you always squash all commits on merge to your main branch? While I appreciate the clean graph that produces, it does make basing new work on unmerged branches very difficult and I suspect leads to problems like what you experienced here with previous changes being lost https://github.com/arkaitzgarro/elastic-apm-laravel/pull/82 if feature branches are not carefully rebased frequently. I suspect that practice works better in projects with small, short-lived feature branches, not the large longer-lived ones I've been working on with you.

I'm happy to adapt to your preferences for this project. I just want to make sure I understand.

README.md
arkaitzgarro: Great, thanks!

arkaitzgarro:
> @arkaitzgarro This should address the outstanding issues. I want to spend some time writing tests. I've not been good about using TDD and need to learn more about codeception.
Great, thank you very much. If you think you are done implementing the necessary changes to support version 7.5 of the agent, I can already give it a try and test it in one of our applications.

> I also updated the stack trace limit default to 50 to be consistent with what it looks like other agents are doing:
> 
> https://docs.google.com/spreadsheets/d/1JJjZotapacA3FkHc2sv_0wiChILi3uKnkwLTjtBmxwU/edit#gid=0

Oh, nice document! Do you know if it's officially maintain by Elastic?

arkaitzgarro:
> > @arkaitzgarro This should address the outstanding issues. I want to spend some time writing tests. I've not been good about using TDD and need to learn more about codeception.

Great, thank you very much. If you think you are done implementing the necessary changes to support version 7.5 of the agent, I can already give it a try and test it in one of our applications.
> 
> > I also updated the stack trace limit default to 50 to be consistent with what it looks like other agents are doing:
> > https://docs.google.com/spreadsheets/d/1JJjZotapacA3FkHc2sv_0wiChILi3uKnkwLTjtBmxwU/edit#gid=0
 
Oh, nice document! Do you know if it's officially maintain by Elastic?

dstepe:
I'm working on some unit tests for the Agent class now. That seems the highest priority for testing since we've added a number of methods. I hope to have that done later today, so hopefully I'll be ready to remove the draft status within a day.

That agent comparison document is linked from the official Elastic agent configuration guide:

https://github.com/elastic/apm/blob/master/specs/agents/configuration.md

The guide (https://github.com/elastic/apm/tree/master/specs/agents) provides a lot of good information and I'm slowing trying to bring the Agent into better alignment. I'm being a little selective since the official PHP agent (https://github.com/elastic/apm-agent-php) is under development now. I don't want to waste my time, but I see value in this package for a while longer.

arkaitzgarro:
> @arkaitzgarro I have couple of questions not directly related to this PR, but not quite worth opening a new issue.

Happy to answer them :)

> 1. What is your expectation for resolving conversations on a PR? I prefer that all conversations be resolved for merging, but I also expect that the person who opened the conversation should be the one to resolve it. That is an indication that the original concern has been addressed. (With the exception of a contributor who is not responding in a timely way and delaying a release.)

Yes, this is also my expectation. If anyone raises a question/concern, I expect that person to be the one resolving it if the original concern has been addressed, or guiding the contributor towards the expected solution. If that wasn't the case in this project please let me know. Also reminders about unanswered/unresolved questions are always welcome. 

> 2. I noticed that the `config/ApmConfigTest.php` is not being run when following your documented development steps for unit tests. Is that intentional and if so why? Or should that test class be moved into the `unit` directory and made part of the standard unit suite?

Oops, this is clearly a mistake. _I think_ the idea was to have the `config` folder outside the `unit` folder to mimic the structure of the source code. Running the command described in the docs will only run unit tests indeed. You could duplicate `unit.suite.yml` file into `config.suite.yml` and run `./vendor/bin/codecept run` to execute both suites. 

> 3. Do you always squash all commits on merge to your main branch? While I appreciate the clean graph that produces, it does make basing new work on unmerged branches very difficult and I suspect leads to problems like what you experienced here with previous changes being lost #82 if feature branches are not carefully rebased frequently. I suspect that practice works better in projects with small, short-lived feature branches, not the large longer-lived ones I've been working on with you.

Yes, squash and merge is my preference to merge commits into `master`, because:

1. Combines all the code changes related to a single logical change (it could be a feature, a bugfix, or an individual change part of a bigger change).
2. If you pick this commit independently from the history, it makes sense on its own. Individual commits are still available in the original PR.
3. Reverting changes is as easy as reverting a single commit. I had terrible situations in the past where GH UI was not available and we needed to revert a feature from the cli. Commit by commit. It was hell and I don't even want to think about it anymore 😅

The issue related to #82 I'm pretty sure is about a wrong `git push` using `--force` on my side. I cannot see the merge commit in the branch, looks like I pushed changes using the force without pulling latest changes.

arkaitzgarro:
> I'm working on some unit tests for the Agent class now. That seems the highest priority for testing since we've added a number of methods. I hope to have that done later today, so hopefully I'll be ready to remove the draft status within a day.

Thanks, I appreciate that. The code coverage is pretty low right now on `master`, so once this PR is merged, I will also make sure our tests are in a better shape before we create the final release.

dstepe:
@arkaitzgarro This should be ready for a review but I see there will be conflicts with at least two of the other PRs in progress. Why don't you wait to merge this one until both of those are merged, then I'll rebase this one and clean it up.

src/Agent.php
```diff
@@ -61,18 +72,48 @@ public function getCollector(string $name): DataCollector
...
+    }
+
+    public function clearCurrentTransaction(): void
+    {
+        $this->current_transaction = null;
+    }
+
+    public function currentTransaction(): Transaction
+    {
+        if (null === $this->current_transaction) {
```
arkaitzgarro: Should we reuse the existing method instead?

```suggestion
        if (!$this->hasCurrentTransaction()) {
```

src/Exception/MissingAppConfigurationException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace AG\ElasticApmLaravel\Exception;
+
+class MissingAppConfigurationException extends \Exception
```
arkaitzgarro: `ElasticApmLaravelException` was created as base exception, but instead we are still extending from `Exception`. Is that intentional?

src/Services/ApmAgentService.php
```diff
@@ -0,0 +1,39 @@
...
+use Psr\Http\Message\RequestInterface;
+
+class ApmAgentService
+{
+    /**
+     * @var Application
+     */
+    protected $app;
+
+    public function __construct(Application $app)
```
arkaitzgarro: Will not injecting the Agent work instead?

src/Agent.php
dstepe: I always feel a little odd negating a method within the class like that. IMHO, it's easier to miss the leading `!` then the logical comparison. So when I'm dealing with class internals, I often just use the raw comparison. But that's my issue and I think makes sense to use the method when stop to consider. This may be a "code smell" that says you need a named method for the other case. Something to think about.

Anyway, I switched to the suggestion.

src/Exception/MissingAppConfigurationException.php
dstepe: Good catch, just a result of copy/paste. It's fixed for both exception classes.

src/Services/ApmAgentService.php
dstepe: I modeled this on the existing ApmCollectorService, which injects some dependencies, but use `App::make()` to get the Agent. It's easy enough to change. Do you want me to also change the ApmCollectorService at the same time?

arkaitzgarro: The `ApmCollectorService` class still needs access to the app to create the instances of the collectors, but we could also inject the Agent directly. I think it's better to inject the Agent, it makes the dependencies of the class very clear.

dstepe: I've updated both service classes to inject the Agent and addressed that change in the tests. The unit tests and a quick real-world test look good. This should be ready for your final consideration for merge.

dstepe:
... except that I now see there are conflicts. I'll get to work rebasing and resolving those.

dstepe:
@arkaitzgarro The rebase is complete and I cleaned up the commits. I believe I resolved all of the conflicts correctly, but double check that your recent merge changes are reflected as you expect.

arkaitzgarro:
> @arkaitzgarro The rebase is complete and I cleaned up the commits. I believe I resolved all of the conflicts correctly, but double check that your recent merge changes are reflected as you expect.

This is great, thank you so much. I have one final question though: why are handling the current transaction in this package, and not in the underlying agent? As far I can see, we are just setting and clearing the value, I think this can be done in the PHP agent.

I'm not saying we should remove it now, but maybe it should be moved to the agent in the next release?

dstepe:
I intend the next major release of the Agent package to assume all responsibility for creating events for reasons like this. It doesn't make sense for every consumer of the Agent to need to collect events, manage parent relationships and more. The agent should be able take on most of that accepting alternate implementations and hooks where appropriate. That's going to a substantial change, and honestly, I'll have to consider how quickly the official agent package is progressing before I decide it's worth doing.

PR #81, opened 2020-10-01
fix/compatibility with agent 75

### LINKED ISSUE #80: Bug: Inconsistent `send` method override in the `Agent` class
Inconsistent `send` method override in the `Agent` class ([child class](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Agent.php#L88), [parent class](https://github.com/nipwaayoni/elastic-apm-php-agent/blob/master/src/Agent.php#L282)):

```
PHP Fatal error:  Declaration of AG\ElasticApmLaravel\Agent::send(): bool must be compatible with Nipwaayoni\Agent::send(): void in /home/vagrant/elastic-apm-laravel/src/Agent.php on line 30
```

countless-integers:
- fix(agent): make send method definition match parent
- chore(composer): make agent version constraint more granular

Closes #80

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "illuminate/routing": "^5.5|^6|^7",
         "illuminate/support": "^5.5|^6|^7",
         "jasny/dbquery-mysql": "^2.0",
-        "nipwaayoni/elastic-apm-php-agent": "^7.4",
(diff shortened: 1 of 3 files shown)


countless-integers:
Just realised that  #78 should fix it too.

PR #83, opened 2020-10-01
chore: use Laravel's runningInConsole method

arkaitzgarro:
Laravel provides a method to check if the app is running in the console or not. No need for manual checks.

### DIFF
--- a/tests/unit/Services/ApmCollectorServiceTest.php
+++ b/tests/unit/Services/ApmCollectorServiceTest.php
@@ -27,7 +27,7 @@ class ApmCollectorServiceTest extends Unit
         $this->configMock = Mockery::mock(Config::class);
         $this->eventsMock = Mockery::mock(Dispatcher::class);
 
-        $this->appMock->expects('runningInConsole')
+        $this->appMock->shouldReceive('runningInConsole')
             ->andReturn(true);
         $this->configMock->shouldReceive('get')
             ->once()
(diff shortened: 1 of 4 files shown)


tests/unit/Services/ApmCollectorServiceTest.php
```diff
@@ -27,6 +27,8 @@ protected function setUp(): void
         $this->configMock = Mockery::mock(Config::class);
         $this->eventsMock = Mockery::mock(Dispatcher::class);
 
+        $this->appMock->shouldReceive('runningInConsole')
```
countless-integers, line 30: With mockery you'll still need `->once()` or `expects()` (it's like `shouldReceive` but with implied once) to set the expectation, which I think you should in this case. E.g.:

```suggestion
        $this->appMock->expects('runningInConsole')
```

countless-integers (changes requested):
[tbh it does the exact same thing with the addition of phpdbg check (which I'm yet to see used by anyone)](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Application.php#L531), but in future versions they added a possibility to set it via env variable for some reason.

arkaitzgarro:
> [tbh it does the exact same thing with the addition of phpdbg check (which I'm yet to see used by anyone)](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Application.php#L531), but in future versions they added a possibility to set it via env variable for some reason.

Computer says no... The annoying part is that if I change the expectation to `times(2)`, it will complain because it's been called only once 😳

countless-integers (approved):
jeez, that's weird. Anyway, back to previous version then. Maybe there's something wrong with mockery setup.

PR #87, opened 2020-10-08
fix(config): default service name value

arkaitzgarro:
The existing `APM_APPNAME` environment variable is getting ignored and replaced by the default config value `defaultServiceName`. To prevent breaking changes on existing applications, I'm setting the default value to the existing value if it exists.

@dstepe could you take a look? The rest of the values look that they are correctly set.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -197,7 +197,7 @@ class ServiceProvider extends BaseServiceProvider
         // Filter out null config options so that the Config class can look for environment variables
         return array_filter(array_merge(
             [
-                'defaultServiceName' => 'Laravel',
+                'defaultServiceName' => config('elastic-apm-laravel.app.appName', 'Laravel'),
                 'frameworkName' => 'Laravel',
                 'frameworkVersion' => app()->version(),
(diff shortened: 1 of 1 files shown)


dstepe:
Ah yes, that makes sense. Sorry I missed that. I tried to account for existing configurations but missed how that would play out. Hopefully that's the only one.

src/ServiceProvider.php
```diff
@@ -197,7 +197,7 @@ protected function getAgentConfig(): array
         // Filter out null config options so that the Config class can look for environment variables
         return array_filter(array_merge(
             [
-                'defaultServiceName' => 'Laravel',
+                'defaultServiceName' => config('elastic-apm-laravel.app.appName', 'Laravel'),
```
countless-integers, line 200: `Laravel` is already the default value in that config.

countless-integers (approved):

src/ServiceProvider.php
arkaitzgarro: Users could remove the default though in their published config. Not harm on having yet another default I guess.

PR #89, opened 2020-10-12
Chore/implement global max trace items

### LINKED ISSUE #85: Max trace items is applied per collector rather than to total events
While writing unit tests for the `Agent` class, I found that the max trace items setting is used to take that number of items from each collector:

https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Agent.php#L70

The README defines max trace items as "Max number of child items displayed when viewing trace details" which is not consistent with the implementation.

Should the implementation be fixed or should the documentation be updated to reflect this is per collector? I vote for fixing the implementation, but makes the behavior rather arbitrary since you make take all of one type and none of another based purely on which collector is processed first.

A better solution might be to create a global `EventCollectionTracker` object which all collectors refer to and increment each time they collect an event. Each collector will then ask that object if it should collect an event and discard the event if not. Only that tracker object would need to be concerned with the max trace items, and the collectors would not consume memory collecting events which won't actually be pushed to APM.

Given that the default is 1000 items, it may be unlikely that anyone ever hits this problem, so maybe it's low priority.

dstepe:
Closes #85 

This took a little more work than I expected, but I think it covers the expected use cases. I'm still looking at EventCollector tests and want to make note about some behaviors in the docs. I wanted to get this out as a draft, though.

I replaced the `CustomCollectorTest` with `EventDataCollectorTest`. The original test did not seem to be actually testing a "custom" collector, just using it as a way to test the abstract `EventDataCollector`. I believe this approach is more explicit and obvious.

The use of the `microtime()` function makes testing difficult, so I added an `EventClock` class which is injected where needed and will provide the time.

The one caveat to how the item limit works relates to when events are added. If a collector were to collect events during the request, but only call `addMeasure()` at the end of the request, the events may not be collected even if they occurred earlier in the transaction. (This is how the original philkra Laravel package handled DB queries for example.) I don't know that anyone will ever run into this, but I want to figure out how to explain it in the docs just in case.


arkaitzgarro:
Looking good, thanks for the fix! I have nothing really to add, looking forward to the PR 👍

dstepe:
I think this is ready. I'd like to spend time on the event collectors, but that's better done in separate PRs.

I see there is a new failing check for styleci, but I get a 404 when I try to view the details for it. What needs to be done for that to pass?

countless-integers:
My spider sense tells me that if you push anything to this branch this check won't be triggered any more. I think @arkaitzgarro disabled it for now.

arkaitzgarro:
> I see there is a new failing check for styleci, but I get a 404 when I try to view the details for it. What needs to be done for that to pass?

Yes, I disabled it. I wanted to know if the Laravel preset was linting variable cases, but it didn't. Looks like an interesting service though, will take a look in the future.

dstepe:
That's a good opportunity fix a typo in the README and clean up the commits. All is good now.

dstepe:
@arkaitzgarro This should be ready now if it looks good to you.

dstepe:
I rebased on to master and fixed a unit test impacted by my changes.

tests/unit/Collectors/JobCollectorTest.php
```diff
@@ -56,14 +58,18 @@ protected function _before()
         $this->agentMock = Mockery::mock(Agent::class);
         $this->configMock = Mockery::mock(Config::class);
 
+        $this->configMock = Mockery::mock(Config::class);
+
         $requestStartTimeMock = Mockery::mock(RequestStartTime::class);
         $requestStartTimeMock->shouldReceive('setStartTime');
         $requestStartTimeMock->shouldReceive('microseconds')->andReturn(1000.0);
 
         $this->collector = new JobCollector(
```
arkaitzgarro, line 70: The merge from master didn't go really well, by bad.

```suggestion
        $eventCounter = new EventCounter();
        $eventClock = new EventClock();

        $this->collector = new JobCollector(
```

arkaitzgarro: Live coding on GitHub, this is a new dimension.

tests/unit/Collectors/JobCollectorTest.php
```diff
@@ -56,14 +58,21 @@ protected function _before()
...
+        $eventCounter = new EventCounter();
+        $eventClock = new EventClock();
+
         $this->collector = new JobCollector(
             $this->app,
             $this->configMock,
-            $requestStartTimeMock
+            $requestStartTimeMock,
+            $eventCounter,
+            $$eventClock
```
arkaitzgarro:
```suggestion
            $eventClock
```

arkaitzgarro (approved):
Changes are looking good, some tests need to be updated though. We added missing tests for collectors, but in this PR the `EventDataCollector` class constructor expects 5 parameters instead of 3. Could you update the broken tests? We can merge it right after.

dstepe:
How about I rebase this branch onto master again and see how it looks? I think that might be better than merging master into the branch. I have a little bit of time now to look at it.

dstepe:
My apologies. I saw the some of the other branches being merged and began making the fixes for the new Collector tests on this branch. That caused the conflicts you experienced. I rebased on to the latest master, cleaned up the tests and force pushed. All the Collector tests are working now.

PR #91, opened 2020-10-16
chore(tests): add tests for collectors

countless-integers:
tests(unit.collectors): fix command collector test
tests(unit.collector): stub for a framework collector test

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,3 +1,5 @@
 <?php
 
+define('LARAVEL_START', 0);
+
 $_SERVER['REQUEST_URI'] = '/path/script.php?var=foo&n=10';
--- a/tests/unit/Collectors/CommandCollectorTest.php
+++ b/tests/unit/Collectors/CommandCollectorTest.php
@@ -63,7 +63,11 @@ class CommandCollectorTest extends Unit
         $requestStartTimeMock = Mockery::mock(RequestStartTime::class);
         $this->configMock = Mockery::mock(Config::class);
 
-        $this->collector = new CommandCollector($this->app, $this->configMock, $requestStartTimeMock);
+        $this->collector = new CommandCollector(
+            $this->app,
+            $this->configMock,
+            $requestStartTimeMock
+        );
         $this->collector->useAgent($this->agentMock);
     }
 
--- a/tests/unit/Collectors/CommandCollectorTest.php
+++ b/tests/unit/Collectors/CommandCollectorTest.php
@@ -50,11 +50,6 @@ class CommandCollectorTest extends Unit
     {
         $this->app = app(Application::class);
         $this->dispatcher = app(Dispatcher::class);
-    }
-
-    protected function setUp(): void
-    {
-        parent::setUp();
 
         $this->commandInputMock = Mockery::mock(InputInterface::class);
         $this->commandOutputMock = Mockery::mock(OutputInterface::class);
@@ -71,7 +66,7 @@ class CommandCollectorTest extends Unit
         $this->collector->useAgent($this->agentMock);
     }
 
-    protected function tearDown(): void
+    protected function _after(): void
     {
         $this->dispatcher->forget(CommandStarting::class);
         $this->dispatcher->forget(CommandFinished::class);
--- a/tests/unit/Collectors/CommandCollectorTest.php
+++ b/tests/unit/Collectors/CommandCollectorTest.php
@@ -21,7 +21,6 @@ class CommandCollectorTest extends Unit
     private const COMMAND_NAME = 'work:do';
     // Use 4 backslashes to match a single backslash: https://stackoverflow.com/a/15369828
     private const COMMAND_IGNORE_PATTERN = '/(?:Application\\\\Commands\\\\DoWork|work:do)/';
-    private const REQUEST_START_TIME = 1000.0;
 
     /** @var Application */
     private $app;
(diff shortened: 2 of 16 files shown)


tests/unit/Collectors/CommandCollectorTest.php
```diff
@@ -111,16 +110,48 @@ public function testCommandFinishedListenerIgnored(): void
...
+        $this->dispatcher->dispatch(
+            new CommandStarting(
+                self::COMMAND_NAME,
+                $this->commandInputMock,
+                $this->commandOutputMock
+            ));
+    }
+
+    /**
+     * @group shit
```
arkaitzgarro: WAT

arkaitzgarro:
@countless-integers I added unit tests for all collectors but `ScheduledTaskCollector`. The events used by this collector are not available in Laravel 5.5 (this is the default version we install for development) and the tests will fail. It would be really nice if we could install different versions of Laravel and run the suite for each version...

tests/unit/Collectors/FrameworkCollectorTest.php
```diff
@@ -0,0 +1,60 @@
...
+    public function testCollectorName(): void
+    {
+        self::assertEquals('framework-collector', $this->collector->getName());
+    }
+
+    public function testItCanRegisterBootingEvent(): void
+    {
+        $this->markTestSkipped('Depending on the order tests are executed, Laravel boot event ');
+
+        $this->app->boot();
```
arkaitzgarro, line 34: Not sure how (I'm assuming `app(Application::class)` follows the singleton pattern), but depending on the order the tests are executed, the application is not [firing the callbacks](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Application.php#L765).

Cannot we make sure `$this->app` is always a new fresh instance?

countless-integers: I've changed it to create a new instance, but I cannot really explain why it was failing before. [Also removed setUp as  it's not a codeception method](https://stackoverflow.com/questions/41934684/what-is-the-difference-between-setup-in-phpunit-and-before-in-codeception), but a phpunit one, so to remove any confusion about when that is run.

tests/bootstrap.php
```diff
@@ -1,3 +1,5 @@
 <?php
 
+define('LARAVEL_START', 0);
```
countless-integers, line 3: 👍 yep, that's a better way of doing it 😉

tests/unit/Collectors/DBQueryCollectorTest.php
```diff
@@ -0,0 +1,143 @@
...
+    /** @var Connection|LegacyMockInterface|MockInterface */
+    private $connectionMock;
+
+    protected function _before(): void
+    {
+        $this->app = app(Application::class);
+        $this->dispatcher = app(Dispatcher::class);
+    }
+
+    protected function setUp(): void
```
countless-integers: why not just use `_before` and `_after` ?

tests/unit/Collectors/DBQueryCollectorTest.php
```diff
@@ -0,0 +1,143 @@
...
+        self::assertGreaterThan(0.0, $measure['start']);
+        self::assertEquals(500.0, $measure['duration']);
+        self::assertEquals('db.mysql.query', $measure['type']);
+        self::assertEquals('query', $measure['action']);
+    }
+
+    public function testDisabledFastQueryExecutedListener(): void
+    {
+        $this->configMock
+            ->shouldReceive('get')
```
countless-integers: if it's meant to be called once then:

```suggestion
            ->expects('get')
```

tests/unit/Collectors/DBQueryCollectorTest.php
```diff
@@ -0,0 +1,143 @@
...
+            ->with('elastic-apm-laravel.spans.querylog.enabled')
+            ->andReturn('auto');
+
+        $this->configMock
+            ->shouldReceive('get')
```
countless-integers:
```suggestion
            ->expects('get')
```

tests/unit/Collectors/DBQueryCollectorTest.php
```diff
@@ -0,0 +1,143 @@
...
+            new QueryExecuted(
+                self::SQL,
+                [1],
+                500,
+                $this->connectionMock
+            )
+        );
+
+        $measure = $this->collector->collect()->first();
+        self::assertNull($measure);
```
countless-integers: would that also mean that `self::assertTrue($this->collector->collect()->isEmpty();`?

tests/unit/Collectors/DBQueryCollectorTest.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public function testFallbackQueryName(): void
+    {
+        $this->configMock
+            ->shouldReceive('get')
```
countless-integers:
```suggestion
            ->expects('get')
```

tests/unit/Collectors/ScheduledTaskCollectorTest.php
```diff
@@ -0,0 +1,44 @@
...
+            $this->app,
+            new Config([]),
+            new RequestStartTime(0.0)
+        );
+        $this->collector->useAgent(Mockery::mock(Agent::class));
+    }
+
+    public function testCollectorName(): void
+    {
+        self::assertEquals('scheduled-task-collector', $this->collector->getName());
```
countless-integers, line 37: 😉

tests/unit/Exception/MissingAppConfigurationExceptionTest.php
```diff
@@ -0,0 +1,31 @@
...
+    private $exception;
+
+    protected function _before(): void
+    {
+        $this->exception = new MissingAppConfigurationException();
+    }
+
+    public function testExceptionMessage(): void
+    {
+        self::assertEquals('Use AgentBuilder::withAppConfig() to provide a configuration object.', $this->exception->getMessage());
```
countless-integers, line 18: I'm surprised linter doesn't complain about line length.

tests/unit/Collectors/FrameworkCollectorTest.php
arkaitzgarro: > I've changed it to create a new instance, but I cannot really explain why it was failing before.

Nice, thanks!

>[Also removed setUp as it's not a codeception method](https://stackoverflow.com/questions/41934684/what-is-the-difference-between-setup-in-phpunit-and-before-in-codeception), but a phpunit one, so to remove any confusion about when that is run.

I'm going to update the rest of tests that are using `setup/teardown` as well.

tests/unit/Collectors/DBQueryCollectorTest.php
arkaitzgarro: Yes, I'm going to update all methods to be `_before` and `_after`.

tests/unit/Collectors/DBQueryCollectorTest.php
arkaitzgarro: This is a cleaner assert, I will change it.

tests/unit/Exception/MissingAppConfigurationExceptionTest.php
arkaitzgarro: [From PSR-2](https://www.php-fig.org/psr/psr-2/):

>There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less.

I love the contradiction, no hard limit but soft limit MUST be.

arkaitzgarro (approved):

PR #93, opened 2020-10-16
chore: run all tests on CI

arkaitzgarro:
Run all Codeception suites, including unit and config tests on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
           command: composer install -n --prefer-dist
       - run:
           name: Run unit tests
-          command: ./vendor/bin/codecept run --xml unit_test_report.xml --coverage
+          command: ./vendor/bin/codecept run --xml test_report.xml --coverage
       - store_test_results:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           command: composer install -n --prefer-dist
       - run:
           name: Run unit tests
-          command: ./vendor/bin/codecept run unit --xml unit_test_report.xml --coverage
+          command: ./vendor/bin/codecept run --xml unit_test_report.xml --coverage
```
countless-integers: should the report still be called unit though?

countless-integers (changes requested):

.circleci/config.yml
arkaitzgarro: Probably not.

```suggestion
          command: ./vendor/bin/codecept run --xml test_report.xml --coverage
```

countless-integers (approved):

countless-integers (approved):

Issue #94, opened 2020-10-16
Change minimum supported Laravel version to 5.6


dstepe:
Original title: Using Log::getLogger() breaks compatibility with older versions

I'm still looking into when the change was made, but it used to be `Log::getMonolog()`. At least in 5.x and possibly 6.x. I'll work on this.

dstepe:
After some more digging, I believe this change occurred in the 5.6 release (I forgot that Laravel did not start using semantic versioning until 6.x, so the 5.5 to 5.6 is a "major" release). The 5.6 release included a major change to the Monolog implementation.

Given that no versions of 5.x are supported for security releases any longer, I am willing to reconsider the minimum supported version for this package. The previous potential breaking change was easy to avoid. This one is not. I suggest that the minimum Laravel version for this package be bumped to 5.6.

arkaitzgarro:
If we are dropping support for 5.5, I would prefer to jump directly into 6. As you mentioned, none of the 5.x versions are supported anymore, I think it makes sense to have the 1.x version of this package compatible with Laravel >=5.5, and the next 2.x version compatible with Laravel >=6.X versions.

dstepe:
I agree. I noticed that the ScheduledTask events will required 6.x as well.

Will you create a release of the 1.x version that will include the new features that are compatible, but limited to Laravel >=5.5 <6.0? I'm going to push to get APM into our apps faster than my dev teams will be able to update (though I'm about to have the conversation regarding just how bad a support situation we are in). I'll need to maintain a compatible fork if there isn't an official release.

arkaitzgarro:
@dstepe your idea will be to support Laravel >=5.5 but using the previous APM agent from philkra right? If that's the case, I can create a branch from the latest 1.X tag, an we can start migrating functionality. I'm afraid it will be an intense manual labour job, non of the latest changes is compatible with the previous version of the agent.

dstepe:
I was actually thinking of branching what we have now, before merging in the change to bump to Laravel 6 minimum. The master branch is working well for me in 5.5+ Laravel apps. I'm not at HEAD with the latest changes, but I think we can get to a working release without too much effort. And leaving that on a branch will allow continued minor releases on the 1.x line. I don't mind putting in the work for that. I'm in a situation in which I want APM for a number of apps sooner than I expect them to be updated to Laravel 6+.

arkaitzgarro:
[...]

PR #95, opened 2020-10-18
Remove Guzzle

veelasky:
Remove explicitly defined guzzle http client adapter.

And also bump laravel version to support version 8.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,8 +21,7 @@
         "illuminate/support": "^5.5|^6|^7",
         "jasny/dbquery-mysql": "^2.0",
         "nipwaayoni/elastic-apm-php-agent": "^7.5",
-        "http-interop/http-factory-guzzle": "^1.0",
-        "php-http/guzzle6-adapter": "^2.0"
+        "http-interop/http-factory-guzzle": "^1.0"
     },
     "require-dev": {
         "codeception/codeception": "^4.1",
@@ -31,6 +30,10 @@
         "friendsofphp/php-cs-fixer": "^2.16",
         "dms/phpunit-arraysubset-asserts": "^0.1.0",
         "codeception/mockery-module": "^0.4.0"
+    },
+    "suggest": {
+        "php-http/guzzle6-adapter": "Guzzle Http client adapter for guzzlehttp/guzzle version 6",
+        "php-http/guzzle7-adapter": "Guzzle Http client adapter for guzzlehttp/guzzle version 7"
     },
     "autoload": {
         "psr-4": {
--- a/README.md
+++ b/README.md
@@ -10,6 +10,22 @@ Elastic APM agent for v2 intake API. Compatible with Laravel 5.5+.
 | --------------------------------------------- | ----------------------------------------------- |
 | <img src="public/apm-transaction-list.png" /> | <img src="public/apm-transaction-detail.png" /> |
 
+## Prerequisites
+
+This package uses `nipwaayoni/elastic-apm-php-agent` which requires you to bring your own Http Client adapter as long as it compliant with [PSR-18](https://www.php-fig.org/psr/psr-18/).
+
+By default, since Laravel version 7 uses `guzzlehttp/guzzle` as the default http client, you need to specify which version of the adapter you want to use.
+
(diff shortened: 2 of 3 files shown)


arkaitzgarro:
Hi @veelasky, could you elaborate what’s the reasoning behind removing the Guzzle adapter?

veelasky:
the underlying `HttpClient` required by `nipwaayoni/elastic-apm-php-agent` only states it need to comply with PSR17 and handled by this package `php-http/discovery`.

Because Guzzle has released version 7, it will be cumbersome if the underlying laravel application uses guzzle version 7, since the adapter for version 6 and 7 is different.

arkaitzgarro:
Shouldn't be the solution to add `php-http/guzzle7-adapter` as well, instead of removing adapters? That way the discovery mechanism will pick up the right adapter based on the installed client. @dstepe does this make sense, or will it cause an issue having both adapters?

dstepe:
I don't think they can both be included. The two Guzzle adapters each require the version of Guzzle they support:

guzzle6-adapter
"guzzlehttp/guzzle": "^6.0"
https://github.com/php-http/guzzle6-adapter/blob/master/composer.json#L21

guzzle7-adapter
"guzzlehttp/guzzle": "^7.0"
https://github.com/php-http/guzzle7-adapter/blob/master/composer.json#L17

That's going to create an immediate dependency conflict.

In general, I advocate that library packages should require the fewest dependencies possible, leaving as much choice to the consumer as they can.

arkaitzgarro:
I see. @veelasky could you add a suggestion in Composer to install `guzzle7-adapter`? And I guess we also need to specify in the README that the final user is responsible to install the adapter.

composer.json
```diff
@@ -15,14 +15,13 @@
     ],
     "require": {
         "php": ">=7.3",
-        "illuminate/database": "^5.5|^6|^7",
-        "illuminate/http": "^5.5|^6|^7",
-        "illuminate/routing": "^5.5|^6|^7",
-        "illuminate/support": "^5.5|^6|^7",
+        "illuminate/database": ">=5.5",
```
arkaitzgarro: Could you do this change in a different PR? This has nothing to do with removing Guzzle. Also, please keep it like `^5.5|^6|^7|^8`, [we are discussing](https://github.com/arkaitzgarro/elastic-apm-laravel/issues/94) which version to support in the new v2.x release, we will address the right composer required format in the consequent PR.

dstepe:
Laravel 7 introduced an HTTP client (https://laravel.com/docs/7.x/http-client) which is based on Guzzle, so the need to for the user of this package to require Guzzle should be limited to Laravel 6.x unless they want a specific version of Guzzle.

composer.json
veelasky: ok

veelasky:
done. 👍 

thanks for the great package, can't wait for the major release.

arkaitzgarro:
Guzzle references are still present in `composer.lock` file though. I guess we need to remove `guzzlehttp/guzzle`, `guzzlehttp/promises` and `guzzlehttp/psr7` right?

veelasky:
oops sorry, thought it was being ignored by `.gitignore`

dstepe:
I believe that most library packages do not include the `composer.lock` file in source code management. Composer only uses that file when you install dependencies for the package project itself, as in when doing development on the package. It won't have any impact when installing the dependencies of a Laravel application, so any mention of Guzzle in the lock file should not have an impact on the Laravel app install.

That said, running a `composer update` with the adapter removed does remove the Guzzle package dependency from the lock file, which could then be committed. Of course, that updates a number of other things, but again, that only impacts people who are developing this package.

I would vote to remove `composer.lock` from the git repo.

arkaitzgarro:
The main issue removing `composer.lock` file is that CI runs out of memory when resolving the dependencies. I wouldn't mind removing it, but circleci [refuses to work](https://app.circleci.com/pipelines/github/arkaitzgarro/elastic-apm-laravel/309/workflows/3fddb0fe-11a4-48fc-ae3f-45bdadb9bc4f/jobs/850).

PR #96, opened 2020-10-18
Fix CommandStartingListener test

dstepe:
I believe this fixes the skipped test:

https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/tests/unit/Collectors/CommandCollectorTest.php#L116


arkaitzgarro:
If tests are passing, it should be fine 👍

PR #97, opened 2020-10-18
Document that job middleware is not called during dispatchNow()

### LINKED ISSUE #90: What is the purpose of job middleware?
I'm not clear on the purpose of job middleware or when I may want to use it.

https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Jobs/Middleware/RecordTransaction.php

The README states it will "record an additional span around your job execution", but does not explain why that would be useful.

The JobCollector appears to do what I want without needing to add the middleware to my job classes.

dstepe:
Closes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ public function middleware()
 }
 ```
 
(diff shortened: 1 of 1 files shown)


arkaitzgarro (approved):
Thanks!

PR #99, opened 2020-10-19
chore: add unit tests for AgentBuilder

arkaitzgarro:


tests/unit/AgentBuilderTest.php
```diff
@@ -0,0 +1,58 @@
...
+
+        self::assertEquals(Agent::class, get_class($agent));
+    }
+
+    public function testBuildAgentWithCollectors()
+    {
+        $this->collector
+            ->shouldReceive('useAgent')
+            ->shouldReceive('getName')
+            ->andReturn('span-collector');
```
countless-integers: I'll sound like a broken record, but I think there should always be a `once` or `expects` to ensure the method is actually called. If it's not called then it's either not needed in the test or something doesn't work according to the expectation:

```suggestion
        $this->collector
            ->shouldReceive('useAgent')
                ->once()
            ->shouldReceive('getName')
                ->twice()
                ->andReturn('span-collector');
```

tests/unit/AgentBuilderTest.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->builder = new AgentBuilder();
+        $this->collector = Mockery::mock(SpanCollector::class);
+    }
+
+    public function testMissingConfigurationException(): void
+    {
+        self::assertThrows(MissingAppConfigurationException::class, function () {
+            $this->builder->build();
+        });
```
countless-integers: This can be done like this, without a need for an extra package:

```suggestion
        $this->expectException(MissingAppConfigurationException::class);
        $this->builder->build();
```

countless-integers (changes requested):

tests/unit/AgentBuilderTest.php
arkaitzgarro: I'm getting old and keep forgetting things like that.

tests/unit/AgentBuilderTest.php
arkaitzgarro: Oh nice, I will remove the extra package then.

countless-integers (approved):

PR #100, opened 2020-10-19
Remove composer.lock and update Laravel supported versions

dstepe:
I've run into composer memory issues in the past and using COMPOSER_MEMORY_LIMIT generally solves the issue. I've not used CircleCI much, but this looks like it might work. Now to see how the build runs.


arkaitzgarro:
We could try. It will make every build way slower, but if it doesn’t crash, I’m ok with it.

dstepe:
I think this is sorted out now. The `composer.lock` file previously included `laravel/framework` version 5.5. I'm still not quite sure what caused that to be included, but it was not installed with a fresh start. That led to a number of test failures because Laravel framework features (like `app()`) were not bootstrapped. That issue was resolved by requiring the `orchestra/testbench` dependency which helps get the Laravel framework running during tests. There were a couple of tests to fix as a result of various updates, but overall, it was straightforward.

I took this opportunity to bump the required Laravel components to a minimum of 6 and a maximum of 8. Version 6 is what you will get for development right now. (More on that with a later PR.)

.circleci/config.yml
```diff
@@ -26,11 +28,13 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - composer-v1-{{ checksum "composer.lock" }}
+            - composer-v1-{{ .Revision }}
```
arkaitzgarro: Could you try using the filename instead of the revision? This works fine for jobs triggered by the same commit, but I don't think it will be really performant between PRs for example, as the revision will constantly change.

```suggestion
            - composer-v1-{{ checksum "composer.json" }}
```

arkaitzgarro (commented):
>The composer.lock file previously included laravel/framework version 5.5. I'm still not quite sure what caused that to be included, but it was not installed with a fresh start. That led to a number of test failures because Laravel framework features (like app()) were not bootstrapped.

Looks like it's been there since the [very beginning](https://github.com/arkaitzgarro/elastic-apm-laravel/commit/d9df7abed1e30ef3819c3af68471bde198a5d0cd#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34). I'm afraid I installed the whole framework instead of separate dependencies, but I forgot to remove it from `composer.lock` file. Strange that it was never removed on `composer update` commands though.

>That issue was resolved by requiring the orchestra/testbench dependency which helps get the Laravel framework running during tests. There were a couple of tests to fix as a result of various updates, but overall, it was straightforward.

Interesting package, thanks for including it. This could even help us testing the package with different versions of Laravel.

>I took this opportunity to bump the required Laravel components to a minimum of 6 and a maximum of 8. Version 6 is what you will get for development right now. (More on that with a later PR.)

Should we bump the minimum version in a different PR then? Specially if you are planning to do more changes.

.circleci/config.yml
```diff
@@ -45,11 +49,13 @@ jobs:
...
             - composer-v1-
-      - run: 
+      - run:
           name: Composer install
           command: composer install -n --prefer-dist
+          environment:
```
arkaitzgarro: You could place all the environment variables on job level instead of command level, same as `CC_TEST_REPORTER_ID`. This makes env variables available for every command.

arkaitzgarro (approved):

.circleci/config.yml
dstepe: I changed this. In theory, we could run into a situation in which the restored `vendor` directory has an old version of a dependency (up to 15 days old). Since we run `composer install`, I believe it would not update if the currently installed version met the version constraint in `composer.json`. Admittedly, this is an edge case and would only occur if a dependency issued a patch release to fix a bug that is impacting our tests. In that scenario, we could update the `composer.json` to indicate the minimum required patch level, though.

.circleci/config.yml
dstepe: I moved these just to keep the environment variables in one place. However, the COMPOSER_MEMORY_LIMIT only applies to `composer` which is only run in that one command in each job, so it won't have any impact on the other commands in the jobs.

dstepe:
I restored the supported versions and will work on a new PR. 

You mentioned a concern with the jobs being slower without the `composer.lock` file. What are your goals for the CI jobs? I ask because this relates to what I want to suggest next.

.circleci/config.yml
arkaitzgarro: I might be wrong, but, does not `composer install` behave like `composer update` if no `composer.lock` file is present? `install` needs to resolve the dependencies first and then will install or update depending on the version prevent in `vendor` folder.

dstepe: I wondered that myself, so I deleted my local `composer.json` but left the vendor directory (simulating what should happen in the job after the cached vendor directory is restored). Running `composer install` resulted in `Nothing to install or update`. I did not try having an intentionally outdated vendor directory and then running install, however.

arkaitzgarro:
> You mentioned a concern with the jobs being slower without the `composer.lock` file. What are your goals for the CI jobs? I ask because this relates to what I want to suggest next.

It's not a big deal for a small project like this one,  but without a `composer.lock` file, now Composer takes around 45s on every job to decide which dependencies to install, while it was 0 before. We are talking about 2 extra minutes to run all CI jobs.

I think it's fine, considering how annoying is to deal with conflicts in the lock file and Composer updating all dependencies when you just want to update one...

.circleci/config.yml
arkaitzgarro: Indeed it installs nothing, but first Composer needs to know what to install to be able to compare with the content of the `vendor` folder. See the difference with [composer.lock](https://app.circleci.com/pipelines/github/arkaitzgarro/elastic-apm-laravel/310/workflows/bdaf9b2c-52c5-43e6-b438-932c89d2a0cc/jobs/852) file and [without](https://app.circleci.com/pipelines/github/arkaitzgarro/elastic-apm-laravel/324/workflows/8ab55e97-c669-4975-995b-5300425be17f/jobs/899). Same console output, completely different execution times :)

PR #101, opened 2020-10-20
Chore/update supported versions

dstepe:
This drops support for Laravel 5.5 and adds support for Laravel 8.x.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,8 @@ jobs:
       - run:
           name: Composer install
           command: composer install -n --prefer-dist
+          environment:
+            COMPOSER_MEMORY_LIMIT: -1
       - save_cache:
           key: composer-v1-{{ checksum "composer.json" }}
           paths:
@@ -35,6 +37,8 @@ jobs:
       - run:
           name: Composer install
           command: composer install -n --prefer-dist
+          environment:
+            COMPOSER_MEMORY_LIMIT: -1
       - run:
           name: Lint PHP code
           command: php ./vendor/bin/php-cs-fixer fix --config .php_cs --dry-run --diff
@@ -55,6 +59,8 @@ jobs:
       - run:
           name: Composer install
           command: composer install -n --prefer-dist
+          environment:
+            COMPOSER_MEMORY_LIMIT: -1
       - run:
           name: Install Code Climate
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   build:
     docker:
@@ -15,8 +15,6 @@ jobs:
       - run:
           name: Composer install
           command: composer install -n --prefer-dist
-          environment:
-            COMPOSER_MEMORY_LIMIT: -1
       - save_cache:
           key: composer-v1-{{ checksum "composer.json" }}
           paths:
@@ -37,15 +35,23 @@ jobs:
       - run:
           name: Composer install
           command: composer install -n --prefer-dist
-          environment:
-            COMPOSER_MEMORY_LIMIT: -1
       - run:
           name: Lint PHP code
           command: php ./vendor/bin/php-cs-fixer fix --config .php_cs --dry-run --diff
 
   unit_tests:
+    parameters:
+      php-version:
+        type: string
+        default: "7.3"
+      laravel-version:
+        type: string
+        default: "6.*"
+      testbench-version:
+        type: string
+        default: "4.*"
     docker:
-      - image: circleci/php:7.3
+      - image: circleci/php:<< parameters.php-version >>
     working_directory: ~/elastic-apm-laravel
     environment:
       CC_TEST_REPORTER_ID: 23639f09f4ef515ddb5fea9baf90f9f5cd9bb5e0801cf1dc062d993a3d70d3eb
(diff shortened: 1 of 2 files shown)


dstepe:
I suggest that part of this PR be the addition of matrix jobs in the CI config to run the tests under multiple versions of PHP as well as both the lowest and highest supported dependency versions. This will add quite a bit to the CircleCI checks, but for a library package, it's a good way to test broad compatibility. Thoughts?

arkaitzgarro:
> I suggest that part of this PR be the addition of matrix jobs in the CI config to run the tests under multiple versions of PHP as well as both the lowest and highest supported dependency versions. This will add quite a bit to the CircleCI checks, but for a library package, it's a good way to test broad compatibility. Thoughts?

Unfortunately, `circleci` does not support matrix jobs. It can be achieved with [different workflows](https://github.com/CircleCI-Public/circleci-demo-workflows/blob/parallel-jobs/.circleci/config.yml), but it's far from pretty. Are you happy with GH actions?

dstepe:
I do like GitHub Actions, but it looks like CircleCI does support matrix jobs now:

https://circleci.com/blog/circleci-matrix-jobs/

Or is that a premium feature?

arkaitzgarro:
> I do like GitHub Actions, but it looks like CircleCI does support matrix jobs now:
> 
> https://circleci.com/blog/circleci-matrix-jobs/
> 
> Or is that a premium feature?

Oh wow, is this feature being available since April? The article says that it's available for all users, I think it's worth to give it a try then.

dstepe:
I use a matrix in the Agent project and really like the coverage it provides. I'll start working on it for CircleCI in this PR.

arkaitzgarro:
@dstepe if this gets more complicated than it should, I'm more than happy to move to GH actions. It looks way easier to combine different PHP and Laravel versions.

https://github.com/arkaitzgarro/elastic-apm-laravel/compare/chore/setup-gh-actions

dstepe:
It's close. The errors above appear to be dependency related. The matrix itself is working, which means moving to GH Actions probably won't make a difference. I'll keep working on this. I think it's close.

arkaitzgarro:
How do you see this working with different Laravel versions? [This is quite neat](https://github.com/arkaitzgarro/elastic-apm-laravel/compare/chore/setup-gh-actions#diff-7314d0ebbd2e9537ae4889316745b4fd2fa43cb86275c9caae18a86ba228b642R15-R19) from GH actions, where you can define dependencies based on other dependencies.

dstepe:
That's a good alternative to what I'm doing now with composer's `--prefer-lowest` option. That only gets the lowest and highest dependency. I'll switch to the `require` technique once I get the various jobs passing.

veelasky:
Since you guys are modifying `composer.json`, I can't help but noticed some of the class importing `Illuminate\Foundation\Application` (example [here](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Services/ApmAgentService.php#L6), [here ](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Collectors/EventDataCollector.php#L9) and [here](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Services/ApmCollectorService.php#L10)) ,  which is in `illuminate\framework` package, and not in `composer.json` file.

May I suggest including `illuminate\contracts` instead? and use the dependency injector to resolve it.

dstepe:
@veelasky Good suggestion. Can you open that as a separate issue? We may choose not to make those changes in this PR since we're focused on the supported versions and pipeline, but we do need address what you pointed out.

veelasky:
will do.. I'll make the issue and send PR for this.

dstepe:
Well, on the plus side, we have a matrix of jobs which covers PHP 7.3 and 7.4 with Laravel versions 6.*, 7.* and 8.*. I think this is a good step forward for now.

However, the CircleCI matrix feature is somewhat limited in how matrixes can be defined (from what I can determine). The version of `orchestra/testbench` must matching the Laravel version, but each matrix parameter can have only a single value. Right now, the `orchestra/testbench` and Laravel versions are provided as separate parameters. Since we know only one combination of each works, we need to use the matrix `exclude` feature to exclude combinations which don't work. Unfortunately, the version of PHP is also part of the matrix, so each exclude item must include all three parameters. The exclude set will quickly become unmanageable.

As I said, I think this works for right now. CircleCI doesn't have a pre-built PHP 8 image that I found and I don't think we need to worry about Laravel 9 until February or so. We can use this pipeline and judge how useful the matrix is. If it looks like something that we want to keep around, maybe we consider GitHub Actions.

dstepe:
For full disclosure, I did think of a couple of alternatives for the matrix parameters.

We could pass in the full package descriptors as the parameter values. For example one value would be `"laravel/framework:6.*" "orchestra/testbench:4.*". That we we only create jobs for matching Laravel and testbench versions. I just think that's ugly, though, but that's a personal opinion.

Another alternative is to use a shell switch statement within the `run` command to set the testbench version based on the Laravel version parameter.

Neither of those options really appeal to me for various reasons, but I thought them worth mentioning.

arkaitzgarro:
I created a [draft](https://github.com/arkaitzgarro/elastic-apm-laravel/pull/106/files) with the same behavior using GH actions. The main difference and I think this is a big plus for GH, is the way `testbench` dependency version is defined. Instead of excluding combinations of dependencies, it is declared in a possitive language, way easier to understand and also maintain. The moment we add PHP8 to circleci, the list of exclusions is gonna explode in complexity.

dstepe:
I'm happy to close this PR in favor of moving straight to GH Actions. I suspect pressure from the developer community will cause CircleCI to catch up eventually, but GH Actions is pretty far out in front on this right now.

arkaitzgarro:
Even though I like circleci, I think in this case GH actions is a better tool for the job. Closing in favour of https://github.com/arkaitzgarro/elastic-apm-laravel/pull/106

PR #107, opened 2020-10-21
chore(deps): remove array subset dependency

arkaitzgarro:
A fancy package that it's not really necessary.


countless-integers (approved):
I wish that assert was part of codeception or phpunit.

arkaitzgarro:
> I wish that assert was part of codeception or phpunit.

it was part of `phpunit`, but removed on version 8 🤷🏻‍♂️

PR #109, opened 2020-10-22
chore: remove check for Laravel 6 class

arkaitzgarro:
Since we only support L6 from now on, there's no need for the class existence check.


countless-integers (approved):
Looks a lot nicer without it.

PR #111, opened 2020-10-22
fix(config): default value for MAXTRACEITEMS env variable

arkaitzgarro:
There was no default value for `MAXTRACEITEMS` env variable, leading to no spans being recorded, as `null` was the default value.


countless-integers (approved):

arkaitzgarro:
@dstepe you might want to cherry pick this commit on 2.x branch.

PR #112, opened 2020-10-22
Support only 5.5 to 5.8

dstepe:
I reduced support on this branch Laravel 5.5 - 5.8. Anyone wanting Laravel 6 or higher should use the 3.x branch.

I added a really basic test of the ServiceProvider to validate the method of fetching the log. That test can be expended to cover more behaviors of the ServiceProvider, but I suggest doing that on the 3.x branch and back-porting to here if we want.

I want to do some more real-world testing before this is merged, even though the tests are passing.


arkaitzgarro:
You can remove the `ScheduledTaskCollector` class and the `RecordTransaction` middleware for jobs, as those are L6 specific.

dstepe:
Closes #94 

I think this is ready. The tests are passing on Laravel versions 5.5 through 5.8, including a new test of the ServiceProvider to make sure getting the Monolog instance works.

I'm not sure why the GitHub actions do not appear on this PR. They are running in my fork and have passed, though.

.github/workflows/run-tests.yml
```diff
@@ -1,6 +1,6 @@
 name: Unit tests
 
-on: push
+on: [push, pull_request]
```
arkaitzgarro, line 3: I added `pull_request` to the list of events that will trigger GH actions, we were missing it.

README.md
```diff
@@ -105,19 +105,6 @@ ApmCollector::startMeasure('my-custom-span', 'custom', 'measure', 'My custom spa
 ApmCollector::stopMeasure('my-custom-span');
 ```
 
-To record an additional span around your job execution, you may include the provided job middleware (Laravel 6+ only https://laravel.com/docs/6.x/queues#job-middleware):
```
arkaitzgarro, line 108: Could you also update the supported Laravel version? It says now 5.5+ which is not really accurate. There's also a mention to Laravel 7, it should be removed as well.

arkaitzgarro (commented):
~I added the event to run GH actions on PRs, but those are failing here and not in your repo... I don't really get it.~

Re-running the jobs solved the issue, computers...

.github/workflows/run-tests.yml
dstepe: Sigh. I should have caught that. It was the end of a long day yesterday. I added it to the lint workflow as well.

README.md
dstepe: Done.

dstepe:
There is something very odd going on with the jobs. As you noted, they fail, but then succeed when re-run. I have observed that on more than one occasion, enough to suspect there is a real problem that needs to be addressed. Again this morning the jobs started by commit failed in your project but passed in mine. Re-running them in yours allowed them to pass.

I'm wondering if there is a cache order problem or something. ~~I see that you are caching the composer cache directory, but not the vendor directory itself. I also see that the cache key in the unit test workflow includes the Laravel version, but the restore key does not. I don't know yet why any of that would make a difference, if it does at all. I'm just making observations at this point.~~

After some more reading on `actions/cache@v2`, the implementation make sense. I'm even more stumped now.

.github/workflows/run-tests.yml
arkaitzgarro: Oh great, thanks.

arkaitzgarro (approved):
Did you test it on real applications, to make a 2.0 release?

arkaitzgarro:
> There is something very odd going on with the jobs. As you noted, they fail, but then succeed when re-run. I have observed that on more than one occasion, enough to suspect there is a real problem that needs to be addressed. Again this morning the jobs started by commit failed in your project but passed in mine. Re-running them in yours allowed them to pass.
> 
> I'm wondering if there is a cache order problem or something. ~I see that you are caching the composer cache directory, but not the vendor directory itself. I also see that the cache key in the unit test workflow includes the Laravel version, but the restore key does not. I don't know yet why any of that would make a difference, if it does at all. I'm just making observations at this point.~
> 
> After some more reading on `actions/cache@v2`, the implementation make sense. I'm even more stumped now.

It makes sense to be a caching problem, because Composer seems to install/load from cache correctly, but somehow some files are missing. If the problem persists, we can disable the cache as first measurement.

dstepe:
@arkaitzgarro Yes, I neglected to mention, but I did run install and run each version of Laravel 5.5 through 5.8 with example app. It hits most of the major features and there were no obvious issues.

PR #113, opened 2020-10-24
chore: use Composer 2 🎉

arkaitzgarro:
Composer 2 is out, and the improvement is just mind-blowing. From 1 minute to 18s for Laravel 6 installation.

Before
Screenshot 2020-10-24 at 14 23 13

After
Screenshot 2020-10-24 at 14 23 40


countless-integers (approved):
suspicious

PR #114, opened 2020-10-26
chore: unit tests for ServiceProvider class

arkaitzgarro:
Some basic tests to verify our services are correctly registered.

### DIFF
--- a/tests/unit/ServiceProviderTest.php
+++ b/tests/unit/ServiceProviderTest.php
@@ -23,6 +23,7 @@ class ServiceProviderTest extends TestCase
     protected function getEnvironmentSetUp($app)
     {
         $app['config']->set('elastic-apm-laravel.app.appName', 'Laravel');
+        $app['config']->set('elastic-apm-laravel.spans.maxTraceItems', EventCounter::EVENT_LIMIT);
     }
 
     public function testRegistersAgent(): void
--- a/tests/unit/ServiceProviderTest.php
+++ b/tests/unit/ServiceProviderTest.php
@@ -1,0 +1,79 @@
+<?php
+
+use AG\ElasticApmLaravel\Agent;
+use AG\ElasticApmLaravel\Collectors\EventCounter;
+use AG\ElasticApmLaravel\Collectors\RequestStartTime;
+use AG\ElasticApmLaravel\Facades\ApmAgent;
+use AG\ElasticApmLaravel\Facades\ApmCollector;
+use AG\ElasticApmLaravel\Middleware\RecordTransaction;
+use AG\ElasticApmLaravel\ServiceProvider;
+use Illuminate\Contracts\Http\Kernel;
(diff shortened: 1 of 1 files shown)


tests/unit/ServiceProviderTest.php
```diff
@@ -0,0 +1,80 @@
...
+        return [
+            ServiceProvider::class,
+        ];
+    }
+
+    protected function getEnvironmentSetUp($app)
+    {
+        $app['config']->set('elastic-apm-laravel.app.appName', 'Laravel');
+        $app['config']->set('elastic-apm-laravel.spans.maxTraceItems', EventCounter::EVENT_LIMIT);
+    }
```
countless-integers, line 27: Those are methods from the testbench package, right?

countless-integers (approved):
not sure if it still qualifies as a unit test, but that's a different topic.

Also all assert methods in PHPUnit, which you use in this test, are static and not dynamic, so it'd be better to call them statically. PHP doesn't complain, sadly.

dstepe:
I was working on similar tests over the last few days and concluded it was not currently worth the effort. The behaviors in the `ServiceProvider` I'm most interested in testing relate to which EventCollectors are registered (depending on http vs console) and the enabled/disabled status of the Agent. Unfortunately, it became very difficult to manage since those behaviors are dependent on code that executes very early and is difficult to influence during testing. I really want to have tests for this, but it quickly became more work than I felt was warranted.

For example, the `testRegistersRequestStartTime` test is easy enough, but it also makes sense to ensure the RequestStartTime is not registered when the Agent is disabled. Unfortunately, testbench only allows you to alter the configuration in the `getEnvironmentSetUp` method, which is called before your test is executed and your `ServiceProvider` is also created before your test is executed. There is no opportunity to change the config for a specific test.

The general consensus sees to be to not unit test `ServiceProvider` classes, but to test the behaviors of your package in real usage. Unfortunately, that's not really how this package works. That implies making http and console requests against the "app" and then testing that the expected outcomes occur, which seems like a lot of non-trivial work. I considered suggesting that we refactor the provider to move some of the work out to helper classes, but I'm right back to questioning if it's really worth the effort.

I hope I'm missing something, but I have been unable to find good alternatives so far.

arkaitzgarro:
Yeap, I faced the same problem when trying to test the different non-happy flows, with different configuration values. And indeed, I couldn't find a solution using testbench.

We could not use testbench for this case and keep the more "traditional" unit test approach, where we create a  `ServiceProvider` instance and call `boot`/`register` methods manually. This will allow us to change the configuration values in different specs, and cover every behaviour of the class.

dstepe:
Trying to create and test a `ServiceProvider` directly was the last thing I tried before deciding I need a break from it. You need to provide an `Application` instance to the constructor, which lead to some challenges. I continue to think testing these behaviors is important, but don't have a good method yet.

PR #119, opened 2020-10-28
3.x Chore/update docs

### LINKED ISSUE #116: The Guzzle6 adapter needs to be >= 2.0
I work with @dstepe and as we were adding this to some existing code that had guzzle6-adapter v 1.1 but I was getting an error, so we discovered that the guzzle6-adapter needs to be > 2.0.

### LINKED ISSUE #117: With AutoDiscovery there is no need to add class to ServiceProvider
I think as of Laravel 5.5 or so, there is no need to add the code to the ServiceProvider, since Laravel now utilizes the command ` @php artisan package:discover` as part of the `"post-autoload-dump"` of the composer json file.

### LINKED ISSUE #118: I was confused by the "Prerequisites" section of the Readme
When I first read the Prerequisites I thought that I had to install `nipwaayoni/elastic-apm-php-agent` but after re-reading I then understood that I just needed to install an HTTP client adapter.  

I am also using a  [Gitlab PHP client](https://github.com/GitLabPHP/Client) and they have a similar requirement,  where they don't provide a client either.  How they describe their installation was more clear for me to understand.

dstepe:
Closes #116 
Closes #117 
Closes #118 

I'll work on updating the 2.x branch README appropriately as well.

### DIFF
--- a/.github/workflows/run-lint.yml
+++ b/.github/workflows/run-lint.yml
@@ -1,6 +1,6 @@
 name: Lint
 
-on: push
+on: [push, pull_request]
 
 jobs:
     lint:
@@ -21,7 +21,7 @@ jobs:
               with:
                   php-version: ${{ matrix.php }}
                   tools: composer:v2
-                  
+
             - name: Get Composer Cache Directory
               id: composer-cache
               run: |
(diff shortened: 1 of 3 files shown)


arkaitzgarro (approved):
Perfect, thank you very much 🙌

PR #120, opened 2020-10-28
2.x Update README regarding prereqs and ServiceProvider discovery

dstepe:
Updates README with some clarifications


arkaitzgarro (approved):
Thanks!

PR #123, opened 2020-10-29
Ensure proper default is used when APM_APPNAME is empty

dstepe:
Closes #122 

I think this addresses the issue described in #122. This behavior results from how the Laravel `config()` helper works and relates to the use of the Agent package default service name. I added an extensive comment in the ServiceProvider to explain what appears to be unnecessarily complex code.

Issue #122, opened 2020-10-29
artisan commands fail with "No service name registered in agent config."


dstepe:
We have observed that artisan commands in the composer post-install scripts fail in our CI/CD pipeline with the error:

```
   Nipwaayoni\Exception\MissingServiceNameException 

  No service name registered in agent config.

  at projects/elastic-apm-php-agent/src/Config.php:105
    101|     {
    102|         $this->resolveLegacyOptions();
    103| 
    104|         if (empty($this->config['serviceName'])) {
  > 105|             throw new MissingServiceNameException();
    106|         }
    107| 
    108|         // TODO validate serviceName matches ^[a-zA-Z0-9 _-]+$
    109|         // TODO support list of server URLs

  1   projects/elastic-apm-php-agent/src/Config.php:70
      Nipwaayoni\Config::validateConfig()

  2   projects/elastic-apm-laravel/src/ServiceProvider.php:109
      Nipwaayoni\Config::__construct(["Laravel", "7.28.4", Object(Monolog\Logger), "error"])
```

This can be reproduced by removing the local `.env` file and running:

```
php artisan package:discover
```
I believe we have identified the source of the problem and I'll be making a PR with a fix later today.

arkaitzgarro:
It might be related to [this random](https://github.com/arkaitzgarro/elastic-apm-laravel/runs/1325515131?check_suite_focus=true#step:7:136) (or not) failing unit test?

dstepe:
I would say there is a very good chance. I'm having a difficult time explaining the "random" part, but I think I understand the cause. I need a little time to confirm and put a change together, but should have something within a day.

arkaitzgarro:
The randomness most probably can be explain by [codeception settings](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/codeception.yml#L4), where we shuffle the test execution, to actually spot situations like that. I can imagine that depending on the order, the configuration is available or not.

Running the tests (2.x branch) with a failing seed makes them always fail:

```
 vendor/bin/codecept run --seed 2103579227
```

dstepe:
[...]

PR #124, opened 2020-10-29
Ensure proper default is used when APM_APPNAME is empty

dstepe:
Related to #122 and PR #123


dstepe:
This should also fix the random unit test failures we've been seeing.

arkaitzgarro (approved):

PR #125, opened 2020-10-30
fix: always initialize an Agent and a Transaction

arkaitzgarro:
We faced a situation when using facades, but having the package disabled with `APM_ACTIVE=false`, was making the application fail. The problem is that we don't initialize an Agent nor start a transaction when the package is not enabled, but, the `ApmAgent` facade expects those to exist, as it exposes functionality related to the Agent and the current transaction.

I made a change to always register the Agent and the middleware but not the collectors if the package is not enabled. This means that an instance of the Agent will always exist, a transaction will be created but no spans will be collected or send to APM.

Edit: there is still a small issue though. We don't create a transaction if it matches the `APM_IGNORE_PATTERNS` pattern, and `ApmAgent::getCurrentTransaction()` will throw an exception as well. I think we should always create a transaction, but don't collect event around it nor send it to APM if it needs to be ignored. @dstepe maybe something the underlying package should take care of?

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -39,10 +39,6 @@ class ServiceProvider extends BaseServiceProvider
         // Always available, even when inactive
         $this->registerFacades();
 
-        if ($this->isAgentDisabled()) {
-            return;
-        }
-
         // Create a single representation of the request start time which can be injected
         // to other classes.
         $this->app->singleton(RequestStartTime::class, function () {
@@ -50,7 +46,10 @@ class ServiceProvider extends BaseServiceProvider
         });
 
         $this->registerAgent();
-        $this->registerCollectors();
+
+        if (!$this->isAgentDisabled()) {
+            $this->registerCollectors();
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


countless-integers (commented):
`registerAgent` also add a a terminate handler that tries to send data from the agent. Does it make sense if collectors are not registered? Maybe we can move that bit elsewhere to not execute code that we know won't do anything.

arkaitzgarro:
> `registerAgent` also add a a terminate handler that tries to send data from the agent. Does it make sense if collectors are not registered? Maybe we can move that bit elsewhere to not execute code that we know won't do anything.

I did nothing because the underlying Agent will not send the events, and I wanted to avoid extra `if` statements everywhere. Also, the agent [will clean up the store](https://github.com/nipwaayoni/elastic-apm-php-agent/blob/master/src/Agent.php#L286), it looks like it's a good idea to actually call `send` :)

countless-integers:
mkay, then send does more than just send 😉

countless-integers (approved):

dstepe:
> then send does more than just send

Unfortunately, yes. That's a remnant of the philkra agent. It will take a major release with breaking changes to sort that out.

countless-integers:
> > then send does more than just send
> 
> Unfortunately, yes. That's a remnant of the philkra agent. It will take a major release with breaking changes to sort that out.

I'm sure well get there eventually 👍

PR #127, opened 2020-11-06
Bug fix: Allow disabling via the config.

cykirsch:
It was broken because array_filter only returns truthy items. When specifying false, it wouldn't pass through the config and would take the upstream's preferred env var instead, which defaults to true.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -222,19 +222,24 @@ class ServiceProvider extends BaseServiceProvider
         }
 
         // Filter out null config options so that the Config class can look for environment variables
-        return array_filter(array_merge(
-            [
-                'defaultServiceName' => $appName,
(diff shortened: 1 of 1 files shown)


arkaitzgarro:
@cykirsch which specific value from the configuration becomes `true` if you specify `false` in your environment?

cykirsch:
When I set APM_ACTIVE to false, the array_filter removes it from the config
given to the agent builder. So the Config uses the ELASTIC_* var instead,
which defaults to true.

arkaitzgarro:
Oh I see it now. Before we were not initialising the Agent when `APM_ACTIVE` was set to false, but now we do so it is available for Facades and they don't fail. 

Thanks for the catch!

arkaitzgarro (approved):

PR #131, opened 2020-11-13
chore(collectors.scheduled-task-collector): handle null exit code from the scheduled task event

### LINKED ISSUE #130: Error: Argument 2 of ScheduledTaskCollector::stopTransaction() must be int
Passing along an error from our app. If it occurs frequently or causes a problem I can look closer, but I don't plan to prioritize it myself right now.

```
Symfony\Component\Debug\Exception\FatalThrowableError: Argument 2 passed to AG\ElasticApmLaravel\Collectors\ScheduledTaskCollector::stopTransaction() must be of the type int, null given, called in /var/www/vendor/arkaitzgarro/elastic-apm-laravel/src/Collectors/ScheduledTaskCollector.php on line 53
```

I don't know why in our case it is null, but I see in the source that `null` is documented as a potential value.

```
// Illuminate\Console\Scheduling\Event

    /**
     * The exit status code of the command.
     *
     * @var int|null
     */
    public $exitCode;
```

countless-integers:
Closes #130

I could not reproduce or find why that would happen...

### DIFF
--- a/src/Collectors/ScheduledTaskCollector.php
+++ b/src/Collectors/ScheduledTaskCollector.php
@@ -66,10 +66,10 @@ class ScheduledTaskCollector extends EventDataCollector implements DataCollector
         );
     }
 
-    protected function stopTransaction(string $transaction_name, int $result): void
+    protected function stopTransaction(string $transaction_name, ?int $result): void
     {
         // Stop the transaction and measure the time
-        $this->agent->stopTransaction($transaction_name, ['result' => $result]);
+        $this->agent->stopTransaction($transaction_name, ['result' => (int) $result]);
         $this->agent->collectEvents($transaction_name);
     }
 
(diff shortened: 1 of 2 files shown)


src/Collectors/ScheduledTaskCollector.php
```diff
@@ -66,10 +66,10 @@ protected function startTransaction(string $transaction_name): Transaction
         );
     }
 
-    protected function stopTransaction(string $transaction_name, int $result): void
+    protected function stopTransaction(string $transaction_name, ?int $result): void
     {
         // Stop the transaction and measure the time
-        $this->agent->stopTransaction($transaction_name, ['result' => $result]);
+        $this->agent->stopTransaction($transaction_name, ['result' => (int) $result]);
```
arkaitzgarro, line 72: If it's null, will it convert to `0`?

countless-integers: yep, not sure what to do with it tbh. or when would it return null. Usually if it returns at all then it means the command worked, because nothing crashed, but I have no idea why it'd happen for @cykirsch .

arkaitzgarro (approved):

PR #133, opened 2020-11-25
[Fix] Add guard clause to ensure return type on getAppConfig

rogercbe:
This PR fixes the following error:
**Return value of AG\ElasticApmLaravel\ServiceProvider::getAppConfig() must be of the type array, null returned**

It happens if the laravel config files are cached and the project is auto-discoverable the call to the config file will return null and the app will crash due to the return type. It can be solved by running "php artisan config:clear", this PR will make the app not to crash if the files are cached.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -249,7 +249,7 @@ class ServiceProvider extends BaseServiceProvider
             $config['appVersion'] = $this->app->make(VersionResolver::class)->getVersion();
         }
 
-        return $config;
+        return empty($config) ? [] : $config;
     }
 
     private function isAgentDisabled(): bool


arkaitzgarro:
@rogercbe do you know if is possible to programatically ask Laravel for fresh config values? Other than calling `Artisan::call('cache:clear')`.  Your fix will solve the issue of the application failing, but it will not be correct, as the configuration values provided by the user will not be applied and it can still be confusing. Better than failing though ;)

dstepe:
I'm afraid I need some help with this issue. My understanding is that configuration caching impacts the `env()` helper, not the `config()` helper. From the Laravel docs:

>Once the configuration has been cached, the .env file will not be loaded and all calls to the env function will return null.

This is why `env()` should only be used in configuration files and the `config()` helper used everywhere else. The `config()` helper must still work correctly or we would see far more problems.

While I'm sure the proposed change fixes the observed symptom, I'm not convinced that we have identified the underlying cause. That implies this may not be the best fix for whatever the real problem is.

I also suggest using a default value in the `config()` call regardless since this would ensure `$config` is an array and prevent array access errors before the return.

```
$config = config('elastic-apm-laravel.app', []);
```

dstepe:
My organization does not cache the config as part of our deployment, so this issue had not come up for me. However, a quick test shows that there is a fundamental problem with the underlying `Agent` and recent move to support configuration through environment variables. I'm not sure if this is the cause of the symptom being addressed by this PR, but it is certainly a real issue.

It appears that when Laravel finds a cached config, it will not parse the `.env` file at all, and therefore the `ELASTIC_APM_` environment variables are not set for the `Agent` to find. I will open a new issue so we can discuss this further. I suggest holding on this PR until we resolve the larger issue.

arkaitzgarro:
@rogercbe we fixed it in a different way in https://github.com/arkaitzgarro/elastic-apm-laravel/pull/135, let us know if it works as expected.

PR #135, opened 2020-12-18
Fix/explicitly configure agent from environment

### LINKED ISSUE #134: ELASTIC_APM_ environment variables not available to the Agent when the Laravel config is cached
When the Laravel application configuration is cached (`artisan config:cache`) the `.env` file is not parsed and no environment variables are set. As a consequence, the underlying `Agent` does not have the expected `ELASTIC_APM_` configuration variables and fails to start.

The `Agent` is following the general practice of other APM agent implementations of configuration through environment variables. The configuration can also be provided as an array, as was previously done. While we wanted to avoid a one-to-one mapping of values through this package to the `Agent`, that may be the only solution to this caching issue.

dstepe:
Closes #134 

I've sat on this for a while, but it's time to get other eyes on it. I can't think of a better way to address this issue than simply explicitly using the `ELASTIC_APM` variables in the `elastic-apm-laravel.php` config file. This will require users to update any previously published local config file if they want to use those variables. I believe the tests show the `APM_` variables continue to work, so I don't believe this introduces any issues. The obvious consequence is that any new features of the `Agent` will need to be added to this package to be available with a cached config.

Once this is merged to the main branch, I'll work on porting it to the 2.x release.

### DIFF
--- a/config/elastic-apm-laravel.php
+++ b/config/elastic-apm-laravel.php
@@ -2,7 +2,7 @@
 
 return [
     // Sets whether the apm reporting should be active or not
-    'active' => env('ELASTIC_APM_ENABLED', env('APM_ACTIVE', true)),
+    'active' => env('APM_ACTIVE', env('ELASTIC_APM_ENABLED', true)),
 
     'log-level' => env('APM_LOG_LEVEL', 'error'),
 
@@ -13,10 +13,10 @@ return [
 
     'app' => [
         // The app name that will identify your app in Kibana / Elastic APM, limited characters allowed
-        'appName' => preg_replace('/[^a-zA-Z0-9 _-]/', '-', env('ELASTIC_APM_SERVICE_NAME', env('APM_APPNAME', 'Laravel'))),
+        'appName' => preg_replace('/[^a-zA-Z0-9 _-]/', '-', env('APM_APPNAME', env('ELASTIC_APM_SERVICE_NAME', 'Laravel'))),
 
         // The version of your app
-        'appVersion' => env('ELASTIC_APM_SERVICE_VERSION', env('APM_APPVERSION')),
+        'appVersion' => env('APM_APPVERSION', env('ELASTIC_APM_SERVICE_VERSION')),
     ],
 
     'env' => [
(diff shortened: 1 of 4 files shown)


config/elastic-apm-laravel.php
```diff
@@ -13,31 +13,36 @@
...
         // Token for x
-        'secretToken' => env('APM_SECRETTOKEN'),
+        'secretToken' => env('ELASTIC_APM_SECRET_TOKEN', env('APM_SECRETTOKEN')),
 
         // Hostname of the system the agent is running on.
         'hostname' => null,
     ],
 
+    'agent' => [
+        'hostname' => env('ELASTIC_APM_HOSTNAME', gethostname()),
```
arkaitzgarro: The `hostname` value is already defined in the `server` block, and it appears correctly in APM, under the `host.hostname` property. Which one should we keep?

config/elastic-apm-laravel.php
```diff
@@ -13,31 +13,36 @@
...
+        'transactionSampleRate' => env('ELASTIC_APM_TRANSACTION_SAMPLE_RATE', 1),
```
arkaitzgarro, line 42: Has this already been implemented 🎉?

config/elastic-apm-laravel.php
```diff
@@ -2,7 +2,7 @@
 
 return [
     // Sets whether the apm reporting should be active or not
-    'active' => env('APM_ACTIVE'),
+    'active' => env('ELASTIC_APM_ENABLED', env('APM_ACTIVE', true)),
```
arkaitzgarro: I have one concern about this implementation. In our documentation we state the following:

>The following options are still supported by this package and will take precedence over their ELASTIC_APM_* counterparts if present.

But with the actual change, it will actually so the opposite right? If both `ELASTIC_APM_*` and `APM_*` variables are present, `ELASTIC_APM_*` variables will take precedence, and might have unexpected consequences for some users.

Could this work instead, or not really?
```suggestion
    'active' => env('APM_ACTIVE', env('ELASTIC_APM_ENABLED', true)),
```

config/elastic-apm-laravel.php
dstepe: I missed that. Keeping the original seems best. I think replacing the `null` with the env value is warranted since the Agent implies you can specify the hostname with that variable. If we don't add it, then there would be no way to explicitly set it when the config is cached.

It makes sense to me to consolidate Agent settings into a specific key in the config array, which is why I added `agent` in there. I know that moving existing items would have consequences and possibly for users to updated previously published config files, so that should wait for a major release.

config/elastic-apm-laravel.php
dstepe: Yes. The transaction sample rate is implemented, though I'm not sure how much real-world usage it's seen yet.

config/elastic-apm-laravel.php
dstepe: You are correct, that would reverse the documented behavior. I don't see why flipping them would be an issue. Thanks for catching that.

dstepe:
Something external to the project appears to have been the cause of the pipeline failures. Can you rerun them on your end? They passed in my fork a few minutes ago.

arkaitzgarro:
> Something external to the project appears to have been the cause of the pipeline failures. Can you rerun them on your end? They passed in my fork a few minutes ago.

Pipelines are green again, merging and releasing 🚀

arkaitzgarro (approved):

PR #148, opened 2021-06-20
Fix `REQUEST_URI` missing and `ServiceProvider` active check

afiqiqmal:
- Fix REQUEST_URI array missing

```
Undefined array key "REQUEST_URI"
```
  Add  null coalescing operator "??" to escape the undefined array key error. `$_SERVER['REQUEST_URI']` not always exist. Occurred when running with Laravel Octane

- Add isAgentDisabled in ServiceProvider boot
Even after disable the status in config, the middleware still running by default. So, by adding APM status check in ServiceProvider to ignore middleware to be executed.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -59,7 +59,11 @@ class ServiceProvider extends BaseServiceProvider
     public function boot(): void
     {
         $this->publishConfig();
-
+        
+        if ($this->isAgentDisabled()) {
+            return;
+        }
+        
         $this->registerMiddleware();
 
         // If not collecting http events, the http middleware will not be executed and an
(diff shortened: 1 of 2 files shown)


countless-integers:
Hi, thanks for contributing! Could you apply linter suggestions to you PR so that all checks can pass? It's looks like it's just whitespace correction.

afiqiqmal:
Im not sure why the lint is not successful. Working fine in my local lint

src/ServiceProvider.php
```diff
@@ -59,7 +59,11 @@ public function register(): void
     public function boot(): void
     {
         $this->publishConfig();
-
+        
+        if ($this->isAgentDisabled()) {
+            return;
+        }
+
```
arkaitzgarro: Some spaces need to be removed still.

```suggestion

        if ($this->isAgentDisabled()) {
            return;
        }

```

arkaitzgarro (approved):

PR #157, opened 2022-02-11
chore: support for Laravel 9

### LINKED ISSUE #156: Laravel 9?
Any plan to upgrade to Laravel 9?

arkaitzgarro:
Add support for Laravel 9, and drop older versions of Laravel (6 and 7) and PHP (7.3 and 7.4).

Tests are passing and I tested it in an application with HTTP requests, commands and scheduled tasks. Seems to be working as expected. It can be tested using the [v4.0.0-BETA-2](https://github.com/arkaitzgarro/elastic-apm-laravel/releases/tag/v4.0.0-BETA-2) tag.

Closes #156

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -8,13 +8,18 @@ jobs:
 
         strategy:
             matrix:
-                php: [8.0, 8.1]
-                laravel: [8.*, 9.*]
+                php: [8.0]
+                laravel: [8.*]
                 include:
-                    - laravel: 9.*
-                      testbench: 7.*
                     - laravel: 8.*
                       testbench: 6.*
+                # php: [8.0, 8.1]
+                # laravel: [8.*, 9.*]
+                # include:
+                #     - laravel: 9.*
+                #       testbench: 7.*
+                #     - laravel: 8.*
+                #       testbench: 6.*
 
         name: PHP${{ matrix.php }} - L${{ matrix.laravel }}
 
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -9,7 +9,7 @@ jobs:
         strategy:
             matrix:
                 php: [8.0, 8.1]
-                laravel: [8.*]
+                laravel: [8.*, 9.*]
                 include:
                     - laravel: 9.*
                       testbench: 7.*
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -14,21 +14,21 @@
         }
     ],
     "require": {
-        "php": ">=7.3",
+        "php": "^8.0.2",
```
countless-integers, line 17: After all these years I still get confused with this syntax :). I thought it should be:

```suggestion
        "php": "^8.0|^8.1",
```

but composer doesn't complain in 8.1 tests, so I guess this is fine. Also, why 8.0.2 specifically?

.github/workflows/run-tests.yml
```diff
@@ -8,15 +8,13 @@ jobs:
 
         strategy:
             matrix:
-                php: [7.4, 7.3]
-                laravel: [8.*, 7.*, 6.*]
+                php: [8.0, 8.1]
+                laravel: [8.*, 9.*]
```
countless-integers, line 12: I've actually realised this weekend that in the `Install dependencies` step we require the laravel package (`composer require "laravel/framework:${{ matrix.laravel }}"`), but the `composer.json` actually specifies individual Illuminate packages. I think installing Laravel works because it would contain all those packages, but we could just limit that step to installing things we require.

composer.json
arkaitzgarro: It is what [Laravel 9 uses as requirement](https://github.com/laravel/framework/blob/9.x/composer.json#L18), and [according to Composer](https://getcomposer.org/doc/articles/versions.md#caret-version-range-), the recommended operator for maximum interoperability when writing library code.

This will allow this package to be installed in PHP `8.0`, `8.1`, `8.2` etc, without the need of updating our requirements.

.github/workflows/run-tests.yml
arkaitzgarro: I think `TestBench` expects the whole Laravel framework to be present right? On our tests we depend on a fully booted application.

I was following what [Spatie does](https://github.com/spatie/laravel-backup/blob/main/.github/workflows/run-tests.yml) in some of their packages.

countless-integers: Oh, interesting, maybe there's more to it indeed.

countless-integers (approved):
🚀

PR #158, opened 2022-03-01
Default log level to INFO

alignwebs:
As per the documentation "The default log level for the agent package is info."

Ref: https://github.com/arkaitzgarro/elastic-apm-laravel#logging

### DIFF
--- a/config/elastic-apm-laravel.php
+++ b/config/elastic-apm-laravel.php
@@ -4,7 +4,7 @@ return [
     // Sets whether the apm reporting should be active or not
     'active' => env('APM_ACTIVE', env('ELASTIC_APM_ENABLED', true)),
 
-    'log-level' => env('APM_LOG_LEVEL', 'error'),
+    'log-level' => env('APM_LOG_LEVEL', 'info'),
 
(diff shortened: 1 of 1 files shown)


arkaitzgarro (approved):

dstepe:
I wanted to point out that "The default log level for the agent package is info." is a true statement. The underlying agent did does default to INFO, though this class defaulted to ERROR prior to this PR. I'm not sure if the documentation should be updated to clarify the distinction between this package and the agent.

arkaitzgarro:
> I wanted to point out that "The default log level for the agent package is info." is a true statement. The underlying agent did does default to INFO, though this class defaulted to ERROR prior to this PR. I'm not sure if the documentation should be updated to clarify the distinction between this package and the agent.

True, I think we should improve the docs instead, looks we are constantly confusing the underlying agent and this package. In the section where we describe the environment variables, we say the default is `error`...

> Log level for the APM Agent package. Must be PSR-3 compliant. Defaults to error.

I'm going to revert the change and think how to make the docs more clear.

PR #162, opened 2022-03-10
Allow ELASTIC_APM_ENVIRONMENT in config

### LINKED ISSUE #161: Standard ELASTIC_APM_ENVIRONMENT value is not used correctly
The package config only looks at `APM_ENVIRONMENT`. PR incoming shortly.

dstepe:
Fixes #161 

Other config values were correctly looking for `APM_` first, then checking `ELASTIC_APM_`. The `environment` configuration was missed.

### DIFF
--- a/config/elastic-apm-laravel.php
+++ b/config/elastic-apm-laravel.php
@@ -24,7 +24,7 @@ return [
         'env' => ['DOCUMENT_ROOT', 'REMOTE_ADDR'],
 
         // Application environment
-        'environment' => env('APM_ENVIRONMENT', 'development'),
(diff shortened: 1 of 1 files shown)


arkaitzgarro (approved):

dstepe:
Thanks @arkaitzgarro. Would you consider making a new release? That would help us so we don't have to change our deployment configuration.

arkaitzgarro:
> Thanks @arkaitzgarro. Would you consider making a new release? That would help us so we don't have to change our deployment configuration.

Yes, I will create a release right away.

PR #172, opened 2022-11-07
fix: schedule task name for closures and invokable objects

### LINKED ISSUE #171: AG\ElasticApmLaravel\Collectors\EventDataCollector::shouldIgnoreTransaction(): Argument #1 ($transaction_name) must be of type string,
```
Php version: 8.1
Package version: v4.0.0
```

I just (re)implemented the package and getting the following exception:

```
TypeError: AG\ElasticApmLaravel\Collectors\EventDataCollector::shouldIgnoreTransaction(): Argument #1 ($transaction_name) must be of type string, null given, called in /data/www/freeontour.com/test/releases/45/vendor/arkaitzgarro/elastic-apm-laravel/src/Collectors/ScheduledTaskCollector.php on line 96
Error
```

Any idea?

arkaitzgarro:
Our implementation to get the transaction name for Scheduled Tasks, was expecting only commands, but it should also accepts closures, invokable objects and jobs. I took the implementation that [Laravel uses to print the scheduled](https://github.com/laravel/framework/blob/61eac9cae4717699ecb3941b16c3d775820d4ca2/src/Illuminate/Console/Scheduling/ScheduleRunCommand.php#L167) name in the terminal. 

The linter was complaining, sorry for the extra diff.

Closes https://github.com/arkaitzgarro/elastic-apm-laravel/issues/171

### DIFF
--- a/src/Collectors/CommandCollector.php
+++ b/src/Collectors/CommandCollector.php
@@ -9,7 +9,6 @@ use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Console\Events\CommandStarting;
 use Illuminate\Support\Facades\Log;
 use Nipwaayoni\Events\Transaction;
-use Throwable;
 
 /**
  * Collects info about ran commands.
@@ -75,7 +74,7 @@ class CommandCollector extends EventDataCollector implements DataCollector
             $this->agent->send();
         } catch (ClientException $exception) {
             Log::error($exception, ['api_response' => (string) $exception->getResponse()->getBody()]);
-        } catch (Throwable $t) {
+        } catch (\Throwable $t) {
             Log::error($t->getMessage());
         }
     }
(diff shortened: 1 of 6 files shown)


src/Collectors/CommandCollector.php
```diff
@@ -75,7 +74,7 @@ protected function send($event): void
             $this->agent->send();
         } catch (ClientException $exception) {
             Log::error($exception, ['api_response' => (string) $exception->getResponse()->getBody()]);
-        } catch (Throwable $t) {
+        } catch (\Throwable $t) {
```
countless-integers, line 77: This looked better before -- is this why the linter complained? Very, very weird.

countless-integers (approved):

src/Collectors/CommandCollector.php
arkaitzgarro: Yes, we were a few versions behind and looks like they [changed the default](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/6662) a couple of weeks ago.

countless-integers: They didn't much of an explanation after I followed the thread to [this comment](https://github.com/symfony/symfony/pull/47709#discussion_r982303782). Must be a Symfony thing.

PR #177, opened 2023-10-11
Add stacktrace to span

### LINKED ISSUE #176: Is there a reason that this agent doesn't allow collecting stacktraces?
First of all, thanks for your work on this repo. We successfully migrated from https://github.com/philkra/elastic-apm-laravel and we love possibility to register our own collectors (DB transactions, Redis, HTTP requests and others).

The only feature that we're missing from the previous APM integration is lack of stacktraces. In APM stacktraces looks as on screenshot and allows us to easily trace the source of the Span
Zrzut ekranu 2023-10-11 o 10 00 24

I found a method that allows us to [set stacktraces on Span in APM agent](https://github.com/nipwaayoni/elastic-apm-php-agent/blob/47c814902449738131c0bd4e457ffef5d8a58718/src/Events/Span.php#L162), so it should be fairly easy to pass it through [Agent](https://github.com/arkaitzgarro/elastic-apm-laravel/blob/master/src/Agent.php#L101-L116)

Summarizing, the question is: Is it intended that this Agent doesn't allow to record spans with stacktraces?

kvas-damian:
Fixes #176

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ The following environment variables are supported in the default configuration:
 |APM_QUERYLOG       | `true` or `false` defaults to 'true'. Set to `false` to completely disable query logging, or to `auto` if you would like to use the threshold feature. |
 |APM_THRESHOLD      | Query threshold in milliseconds, defaults to `200`. If a query takes longer then 200ms, we enable the query log. Make sure you set `APM_QUERYLOG=auto`. |
 |APM_MAXTRACEITEMS  | Defaults to `1000`. Max number of child items displayed when viewing trace details. |
+|APM_RENDERSOURCE   | Defaults to `true`. Include source code in query span. |
 
 You may also publish the `elastic-apm-laravel.php` configuration file to change additional settings:
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ The following environment variables are supported in the default configuration:
 |APM_QUERYLOG       | `true` or `false` defaults to 'true'. Set to `false` to completely disable query logging, or to `auto` if you would like to use the threshold feature. |
 |APM_THRESHOLD      | Query threshold in milliseconds, defaults to `200`. If a query takes longer then 200ms, we enable the query log. Make sure you set `APM_QUERYLOG=auto`. |
 |APM_MAXTRACEITEMS  | Defaults to `1000`. Max number of child items displayed when viewing trace details. |
-|APM_RENDERSOURCE   | Defaults to `true`. Include source code in query span. |
+|APM_RENDERSOURCE   | Defaults to `false`. Include source code in query span. |
 
 You may also publish the `elastic-apm-laravel.php` configuration file to change additional settings:
 
--- a/config/elastic-apm-laravel.php
+++ b/config/elastic-apm-laravel.php
@@ -54,7 +54,7 @@ return [
         'maxTraceItems' => env('APM_MAXTRACEITEMS', 1000),
 
         // Depth of backtraces
-        'backtraceDepth' => env('APM_BACKTRACEDEPTH', env('ELASTIC_APM_STACK_TRACE_LIMIT', 25)),
+        'backtraceDepth' => env('APM_BACKTRACEDEPTH', env('ELASTIC_APM_STACK_TRACE_LIMIT', 0)),
 
         'querylog' => [
             // Set to false to completely disable query logging, or to 'auto' if you would like to use the threshold feature.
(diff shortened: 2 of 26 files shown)


README.md
```diff
@@ -88,6 +88,7 @@ The following environment variables are supported in the default configuration:
 |APM_QUERYLOG       | `true` or `false` defaults to 'true'. Set to `false` to completely disable query logging, or to `auto` if you would like to use the threshold feature. |
 |APM_THRESHOLD      | Query threshold in milliseconds, defaults to `200`. If a query takes longer then 200ms, we enable the query log. Make sure you set `APM_QUERYLOG=auto`. |
 |APM_MAXTRACEITEMS  | Defaults to `1000`. Max number of child items displayed when viewing trace details. |
+|APM_RENDERSOURCE   | Defaults to `true`. Include source code in query span. |
```
arkaitzgarro: I think the default value should be `false`, and make it an opt in feature. Adding a stacktrace for every single span will generate a significant amount of extra data on Elastic documents, and this can be an issue for applications with significant traffic.

src/Helpers/StacktraceExtractor.php
```diff
@@ -0,0 +1,81 @@
...
+                'pre_context' => optional($sourceCode->get('pre_context'))->toArray(),
+                'context_line' => optional($sourceCode->get('context_line'))->first(),
+                'post_context' => optional($sourceCode->get('post_context'))->toArray(),
+            ];
+        })->values()->toArray();
+    }
+
+    private static function getSourceCode(array $stackTrace, Config $config): Collection
+    {
+        if (false === $config->get('elastic-apm-laravel.spans.renderSource', true)) {
```
arkaitzgarro: Can you move this to the beginning of `getStacktrace` method? This way we avoid unnecessary loops.

arkaitzgarro (commented):
Thanks for this, I will take a look as soon as possible and see if we can add some tests.

config/elastic-apm-laravel.php
```diff
@@ -63,5 +63,8 @@
             // If a query takes longer then 200ms, we enable the query log. Make sure you set enabled = 'auto'
             'threshold' => env('APM_THRESHOLD', 200),
         ],
+
+        // Render source code in stack traces
+        'renderSource' => env('APM_RENDERSOURCE', true),
```
arkaitzgarro: Can you move this into the `spans` configuration block? The code is trying to access `elastic-apm-laravel.spans.renderSource` property.

arkaitzgarro:
Hey @kvas-damian,

I managed to make it work locally, and it shows the stacktraces on span level.

Screenshot 2023-10-22 at 11 20 39

Could you fix the following:
- Move `renderSource` config into `span` block, since the code expects it to be there. And make it `false` by default, I'm a bit concern it may blow up some Kibana clusters with a high number of spans. Update the docs.
- Move `getStacktrace` to the parent method, to avoid extra loops.

kvas-damian:
Thanks for your feedback. I'll try to apply changes today

config/elastic-apm-laravel.php
kvas-damian: it already is in `spans`array ;)

tests/config/ApmConfigTest.php
```diff
@@ -59,7 +59,7 @@ public function testDefaultValues()
         $this->assertEquals(1000, $this->config['spans']['maxTraceItems']);
         $this->assertEquals(25, $this->config['spans']['querylog']['enabled']);
         $this->assertEquals(200, $this->config['spans']['querylog']['threshold']);
-        $this->assertEquals(25, $this->config['spans']['backtraceDepth']);
+        $this->assertEquals(0, $this->config['spans']['backtraceDepth']);
```
arkaitzgarro, line 62: Can you also add an assert for `renderSource` default value?

config/elastic-apm-laravel.php
arkaitzgarro: Oh I see 😅

arkaitzgarro:
Looking good, thanks @kvas-damian !

kvas-damian:
Unfortunately, I wasn't able to find a way to cover `StacktraceExtractor`... It depend on `base_path()` from Laravel, which cannot be run without `bootstrap/app.php` :(

tests/config/ApmConfigTest.php
kvas-damian: https://github.com/arkaitzgarro/elastic-apm-laravel/pull/177/commits/a4fe1f83a011abaa29d19d730a8c642027148ad4

claspina:
What is missing to merge this? Maybe I can help

arkaitzgarro (approved):

PR #178, opened 2023-10-24
chore: enable PHP 8.2 CI jobs

arkaitzgarro:
Run linters and tests in PHP 8.2. I also fixed some lint problems related to optional paramenters.

### DIFF
--- a/src/Jobs/Middleware/RecordTransaction.php
+++ b/src/Jobs/Middleware/RecordTransaction.php
@@ -9,10 +9,7 @@ class RecordTransaction
     /**
      * Wrap the job processing in an APM transaction.
      *
-     * @param mixed    $job
      * @param callable $next
-     *
-     * @return mixed
      */
     public function handle($job, $next)
     {
--- a/src/Jobs/Middleware/RecordTransaction.php
+++ b/src/Jobs/Middleware/RecordTransaction.php
@@ -8,10 +8,8 @@ class RecordTransaction
 {
     /**
      * Wrap the job processing in an APM transaction.
-     *
-     * @param callable $next
      */
-    public function handle($job, $next)
+    public function handle(mixed $job, callable $next)
     {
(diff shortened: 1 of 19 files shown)


src/Jobs/Middleware/RecordTransaction.php
```diff
@@ -9,10 +9,7 @@ class RecordTransaction
     /**
      * Wrap the job processing in an APM transaction.
      *
-     * @param mixed    $job
      * @param callable $next
-     *
-     * @return mixed
      */
     public function handle($job, $next)
```
countless-integers: `mixed` type was made available in PHP8.0 which is the min supported version, so we could add it as a type here along with callable and get rid of phpdocs for types.

src/Services/ApmCollectorService.php
```diff
@@ -44,10 +44,10 @@ public function __construct(Application $app, Dispatcher $events, Config $config
     public function startMeasure(
         string $name,
         string $type = 'request',
-        ?string $action = null,
-        ?string $label = null,
-        ?float $start_time = null
```
countless-integers, line 49: since this is a public function's declaration change it could potentially break compatibility for somebody that was passing `null` here. Unlikely, but just mentioning.

src/Services/ApmCollectorService.php
```diff
@@ -82,7 +82,7 @@ public function addCollector(string $collector_class): void
         );
     }
 
-    public function captureThrowable(\Throwable $thrown, array $context = [], ?Transaction $parent = null)
+    public function captureThrowable(\Throwable $thrown, array $context = [], Transaction $parent = null): void
```
countless-integers: Same comment as previous one for that `?`

src/Services/ApmCollectorService.php
arkaitzgarro: Why did the linter remove the `?` then? Do you think we should override the rule?

countless-integers (approved):
I noticed test classes have no namespaces, but would be a different PR.

PR #180, opened 2024-03-21
feat: make package compatible with Laravel 11

arkaitzgarro:
Based on https://github.com/arkaitzgarro/elastic-apm-laravel/pull/179 but with [PHP 8.2 fixes](https://github.com/arkaitzgarro/elastic-apm-laravel/pull/178).

Closes https://github.com/arkaitzgarro/elastic-apm-laravel/pull/179

### DIFF
--- a/tests/unit/Collectors/JobCollectorTest.php
+++ b/tests/unit/Collectors/JobCollectorTest.php
@@ -36,16 +36,16 @@ class JobCollectorTest extends Unit
      */
     private $collector;
 
-    /** @var Agent|\Mockery\LegacyMockInterface|\Mockery\MockInterface */
+    /** @var Agent|Mockery\LegacyMockInterface|Mockery\MockInterface */
     private $agentMock;
 
-    /** @var Job|\Mockery\LegacyMockInterface|\Mockery\MockInterface */
+    /** @var Job|Mockery\LegacyMockInterface|Mockery\MockInterface */
     private $jobMock;
 
-    /** @var Transaction|\Mockery\LegacyMockInterface|\Mockery\MockInterface */
+    /** @var Transaction|Mockery\LegacyMockInterface|Mockery\MockInterface */
     private $transactionMock;
 
-    /** @var Config|\Mockery\LegacyMockInterface|\Mockery\MockInterface */
+    /** @var Config|Mockery\LegacyMockInterface|Mockery\MockInterface */
     private $configMock;
 
     protected function _before()
--- a/tests/unit/Collectors/JobCollectorTest.php
+++ b/tests/unit/Collectors/JobCollectorTest.php
@@ -15,7 +15,6 @@ use Illuminate\Queue\Events\JobFailed;
 use Illuminate\Queue\Events\JobProcessed;
 use Illuminate\Queue\Events\JobProcessing;
 use Illuminate\Queue\Jobs\SyncJob;
-use Illuminate\Support\Facades\Context;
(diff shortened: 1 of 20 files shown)


countless-integers:
`tests/unit/Collectors/JobCollectorTest.php:testJobProcessingListenerIgnored` is failing and looks like it might need some looking into.

tests/unit/Collectors/JobCollectorTest.php
```diff
@@ -104,6 +104,7 @@ public function testJobProcessingListenerIgnored()
     {
         $this->patternConfigReturn(self::JOB_IGNORE_PATTERN);
         $this->jobMock->shouldReceive('resolveName')->once()->andReturn(self::JOB_NAME);
+        $this->jobMock->shouldReceive('payload');
```
arkaitzgarro: @countless-integers  I added this to make the test pass, I guess L11 introduced a new internal method? The thing is, if I do `$this->jobMock->shouldReceive('payload')->once();` locally it fails:

```
  Method payload(<Any Arguments>) from Mockery_0_Illuminate_Contracts_Queue_Job should be called
   exactly 1 times but called 0 times.
```

Not sure if I'm doing the right thing 🤷🏻‍♂️

countless-integers: [The method exists in L10 too](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Contracts/Queue/Job.php#L26). Not sure if I'm reading the test right, but it seems like it's checking if the job would be ignored based on the name pattern and the fact that `payload` is called, to me, indicates that it's not.

countless-integers: [I don't see anything in the upgrade guide on Job or Worker changes](https://laravel.com/docs/11.x/upgrade#queues). All the other tests use the same pattern, the only difference is how this is implemented for processing -- in others get name is called first, but in processing timer is started first. 

But if it passes if you add the mock then it could be that there's something else trying to process the job first. Not sure why would that happen in a unit test though.

countless-integers: >   Method payload(<Any Arguments>) from Mockery_0_Illuminate_Contracts_Queue_Job should be called
>   exactly 1 times but called 0 times.

Any chance you're not testing with Laravel 11 locally?

arkaitzgarro: > [The method exists in L10 too](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Contracts/Queue/Job.php#L26). Not sure if I'm reading the test right, but it seems like it's checking if the job would be ignored based on the name pattern and the fact that `payload` is called, to me, indicates that it's not.

This is the reason the test is failing, it was added on L11.

https://github.com/laravel/framework/blob/11.x/src/Illuminate/Log/Context/ContextServiceProvider.php#L39

tests/unit/Collectors/JobCollectorTest.php
```diff
@@ -107,6 +107,12 @@ public function testJobProcessingListenerIgnored()
         $this->agentMock->shouldNotReceive('startTransaction');
         $this->agentMock->shouldNotReceive('getTransaction');
 
+        // For Laravel 11+
+        if (class_exists('\Illuminate\Support\Facades\Context')) {
```
countless-integers, line 111: Ah yeah, that's a new thing in L11.

countless-integers (approved):

PR #182, opened 2024-04-16
chore: remove compatibility with older versions of Laravel

arkaitzgarro:
Versions 8 and 9 of Laravel are not supported anymore. Remove PHP 8.0 as well.


countless-integers (approved):
In this case, I think you can also drop [v6 and v7 of `orchestra/testbench`](https://packages.tools/testbench.html#version-compatibility) from the `require-dev`.

PR #187, opened 2025-05-28
feat: add support for Laravel 12

arkaitzgarro:
Add compatibility with Laravel 12.x and drop Laravel 10.x

### DIFF
--- a/.github/workflows/run-lint.yml
+++ b/.github/workflows/run-lint.yml
@@ -1,8 +1,12 @@
 name: Lint
 
-on: 
-  - push
-  - pull_request
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
 
 jobs:
   lint:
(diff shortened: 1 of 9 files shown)


.github/workflows/run-lint.yml
```diff
@@ -1,8 +1,12 @@
 name: Lint
 
-on: 
-  - push
-  - pull_request
+on:
```
arkaitzgarro, line 3: This will make actions to run only once.

src/Agent.php
```diff
@@ -44,7 +44,7 @@ public function __construct(
         Connector $connector,
         EventFactoryInterface $eventFactory,
         TransactionsStore $transactionsStore,
-        Repository $app_config
+        Repository $app_config,
```
arkaitzgarro, line 47: No idea why this keeps coming every now and then.

countless-integers (approved):

countless-integers:
Since we're dropping support then the release should be a breaking change ofc

PR #191, opened 2026-06-01
chore: remove support for PHP 8.2 and Laravel 11

arkaitzgarro:
- Laravel 11 is not supported anymore, and PHP 8.2 security support will end in December 2026. Time to remove them.
- Code Climate integration seems to be broken, I disabled it but coverage is still shown in CI output.
- This means a breaking change and major version bump on next release.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -54,7 +54,7 @@ jobs:
           composer update --prefer-dist --no-interaction --no-suggest
 
       - name: Execute tests
-        uses: paambaati/codeclimate-action@v2.7.4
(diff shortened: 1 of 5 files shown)


.github/workflows/run-tests.yml
```diff
@@ -56,10 +54,4 @@ jobs:
           composer update --prefer-dist --no-interaction --no-suggest
 
       - name: Execute tests
-        uses: paambaati/codeclimate-action@v2.7.4
```
arkaitzgarro, line 59: Tests were failing, seems like code climate stopped working. Showing the coverage in CI should be more than ok.

countless-integers (approved):

